Issue #46, opened 2020-03-25
pretrained-model load ERROR!


HK017:
OP_REQUIRES failed at save_restore_v2_ops.cc:184 : Not found: Key box_net/box-0-bn-3/beta/Momentum not found in checkpoint

pre-training model loading error，please help me，thanks a lot!!!

HK017:
my tensorflow is 1.15.2 or 2.1.0 , both tf versions reporte this error.

fsx950223:
> my tensorflow is 1.15.2 or 2.1.0 , both tf versions reporte this error.

The model only support inference. You could download efficientnet pretrained ckpt as backbone ckpt.

Ely-S:
I can repro this. This is the full error for reference:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/tensorflow/python/client/session.py", line 1365, in _do_call
    return fn(*args)
  File "/usr/local/lib/python3.7/dist-packages/tensorflow/python/client/session.py", line 1350, in _run_fn
    target_list, run_metadata)
  File "/usr/local/lib/python3.7/dist-packages/tensorflow/python/client/session.py", line 1443, in _call_tf_sessionrun
    run_metadata)
tensorflow.python.framework.errors_impl.NotFoundError: From /job:worker/replica:0/task:0:
Key box_net/box-0-bn-3/beta/Momentum not found in checkpoint
	 [[{{node save/RestoreV2}}]]
```

From inspecting the training checkoint, `box_net/box-0-bn-3/beta/Momentum` is definitely not in the checkpoint, so the code is working fine.

The issue seems to be, like @fsx950223 said, that the checkpoints only support inference, not training.

roadcode:
@Ely-S  you can use the efficientnet pretrained ckpt [https://github.com/tensorflow/tpu/tree/master/models/official/efficientnet](url) and use the --backbone_ckpt param to restore the checkpoint

HK017:
@fsx950223 @Ely-S @roadcode  Thanks a lot,  I have solved my problem

PR #49, opened 2020-03-25
Train from public checkpoints

Ely-S:
Hi,

I wanted to make it easy to start training from one of the publicly released checkpoints instead of having to train from scratch. This branch adds a `--ckpt` argument to main.py to do that. It's like `--backbone_ckpt` but for the whole model.

It works like this:

```
python3 main.py \
  --training_file_pattern="$DATA_DIR/*.tfrecord" \
  --model_name=$MODEL \
  --ckpt=$CHECKPOINT_PATH \
  --model_dir=$MODEL_DIR
```

I'd appreciate feedback.

Addresses issue #46


efficientdet/det_model_fn.py
```diff
@@ -513,18 +513,35 @@ def metric_fn(**kwargs):
...
-      logging.info('restore variables from %s', params['backbone_ckpt'])
-      if params['ckpt_var_scope'] is None:
-        ckpt_scope = params['backbone_name']  # Use backbone name in default.
-      else:
-        ckpt_scope = params['ckpt_var_scope']
-      tf.train.init_from_checkpoint(
-          params['backbone_ckpt'],
-          utils.get_ckt_var_map(params['backbone_ckpt'], ckpt_scope + '/',
-                                params['backbone_name'] + '/'))
+      tf.logging.info('restore variables from %s' % checkpoint)
```
mingxingtan: Could you use logging.info instead of tf.logging.info?  They are equivalent, but tf.logging is deprecated in tf2

Ely-S: sure

Ely-S:
I'm glad to help!

PR #95, opened 2020-03-29
Soft nms

### LINKED ISSUE #89: Today the model detects less object
Yesterday, i run the script with previous pretained ckpt and source code(seems created on 3/16)on the test image which its result is the same as demo, but today i update the Pretrained EfficientDet Checkpoints( created on 3/27)and source code, and rerun the script, the result shows there is less objects are detected. min_score_thresh both equals 0.2, max_boxes_to_draw both equals 100. Why? 

above image inferenced by driver.serve_images()

above image inferenced by driver.inference()
demo
above image is demo image

fsx950223:
May fix #89

### DIFF
--- a/efficientdet/inference.py
+++ b/efficientdet/inference.py
@@ -145,7 +145,8 @@ def restore_ckpt(sess, ckpt_path, enable_ema=True, export_ckpt=None):
 
 
 def det_post_process(params: Dict[Any, Any], cls_outputs: Dict[int, tf.Tensor],
-                     box_outputs: Dict[int, tf.Tensor], scales: List[float]):
+                     box_outputs: Dict[int, tf.Tensor], scales: List[float],
+                     min_score_thresh=0.2, max_boxes_to_draw=50):
   """Post preprocessing the box/class predictions.
 
   Args:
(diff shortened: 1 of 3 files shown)


fsx950223:
image
I got above image after sort boxes order by scores.

fsx950223:
As ckpt has changed, the demo image in README should be changed.

efficientdet/inference.py
```diff
@@ -477,11 +495,13 @@ def inference(self,
           self.model_name, images, **self.params)
       restore_ckpt(sess, self.ckpt_path, enable_ema=True, export_ckpt=None)
       # for postprocessing.
-      params.update(dict(batch_size=len(raw_images), disable_pyfun=False))
+      params.update(dict(batch_size=len(raw_images), disable_pyfun=True))
```
mingxingtan: Could you first set it to False. I will first merge your PR, and verify that disable_pyfun=True/False have the same outputs. After that, I will change the default value to True.

mingxingtan: Another choice is to have a class member as disable_pyfun and users can choose to set different values.

efficientdet/inference.py
```diff
@@ -491,11 +511,6 @@ def inference(self,
         classes = output_np[:, 6].astype(int)
         scores = output_np[:, 5]
 
-        # This is not needed if disable_pyfun=True
-        # convert [x, y, width, height] to [ymin, xmin, ymax, xmax]
-        # TODO(tanmingxing): make this convertion more efficient.
```
mingxingtan: You can guard this line with the condition of whether disable_pyfun=True or False.

fsx950223:
cc @mingxingtan

mingxingtan:
Thank you @fsx950223 . The result looks great!  I wonder is this soft nms faster?

fsx950223:
> Thank you @fsx950223 . The result looks great! I wonder is this soft nms faster?

They have similar performance, depend on the situation.

PR #104, opened 2020-03-30
add export tflite script

fsx950223:
I export the tflite model by the script on ```tf-nightly```.
But core dump when execute ```invoke```.

### DIFF
--- a/efficientdet/export_tflite.py
+++ b/efficientdet/export_tflite.py
@@ -1,0 +1,75 @@
+# Copyright 2020 Google Research. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


dreamibor:
is this the problem of TensorFlow or the problem of how the model is extracted using saved model? Will converting from protobuf helps?

BeautyMess:
While converting to tflite using Tensorflow 2.1, I got the following error:

`Exception: <unknown>:0: error: loc("cond/while/LoopCond@_functionalize_if_else_branch_3"): operand type tensor<2x!tf.resource<tensor<*xi32>>> is incompatible with result type tensor<2x!tf.resource> at index 8
<unknown>:0: error: loc("cond_1/while/LoopCond@_functionalize_if_else_branch_4"): operand type tensor<2x!tf.resource<tensor<*xi32>>> is incompatible with result type tensor<2x!tf.resource> at index 8
<unknown>:0: error: loc("cond_2/while/LoopCond@_functionalize_if_else_branch_5"): operand type tensor<2x!tf.resource<tensor<*xi32>>> is incompatible with result type tensor<2x!tf.resource> at index 8
<unknown>:0: error: loc("cond_3/while/LoopCond@_functionalize_if_else_branch_6"): operand type tensor<2x!tf.resource<tensor<*xi32>>> is incompatible with result type tensor<2x!tf.resource> at index 8
`
If I have any process, I will update them here. And thanks for your great work anyway

PR #129, opened 2020-04-01
let ServingDriver support eager execution

fsx950223:


efficientdet/inference.py
```diff
@@ -335,36 +338,32 @@ def build(self,
...
-        box_outputs,
-        scales,
-        min_score_thresh=min_score_thresh,
-        max_boxes_to_draw=max_boxes_to_draw)
-
     if not self.sess:
       self.sess = tf.Session()
-    restore_ckpt(self.sess, self.ckpt_path, enable_ema=True, export_ckpt=None)
+    with self.sess.graph.as_default():
+      image_files = tf.placeholder(tf.string, name='image_files', shape=[None])
```
mingxingtan, line 344: Can tf.placeholder still work for eager mode?

fsx950223: I will work in graph context. As graph context will disable eager execution.
For example:
```python
with tf.Graph() as g:
    tf.placeholder()
```
but it will raise error use it directly in eager mode.
With the above change, I could use ServingDriver in another script which enables eager execution.

PR #190, opened 2020-04-09
add tpu args

### LINKED ISSUE #181: Does GPU have to use TpuBatchNormalization when training model?
https://github.com/google/automl/blob/fe439281c367b742dc2358fa9d13c55991b7dfda/efficientdet/utils.py#L218-L219

fsx950223:
Increase training speed about 30% per gpu.
Screenshot from 2020-04-09 12-54-59
Fix #181

### DIFF
--- a/efficientdet/efficientdet_arch.py
+++ b/efficientdet/efficientdet_arch.py
@@ -78,7 +78,8 @@ def remove_variables(variables, resnet_depth=50):
 def resample_feature_map(feat, name, target_width, target_num_channels,
                          apply_bn=False, is_training=None,
                          conv_after_downsample=False,
-                         use_native_resize_op=False, pooling_type=None):
+                         use_native_resize_op=False, pooling_type=None,
+                         use_tpu=False):
   """Resample input feature map to have target number of channels and width."""
 
   _, width, _, num_channels = feat.get_shape().as_list()
@@ -99,6 +100,7 @@ def resample_feature_map(feat, name, target_width, target_num_channels,
             is_training_bn=is_training,
             relu=False,
             data_format='channels_last',
+            use_tpu=use_tpu,
             name='bn')
     return feat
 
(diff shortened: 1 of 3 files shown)


efficientdet/efficientdet_arch.py
```diff
@@ -401,6 +409,7 @@ def build_feature_network(features, config):
             conv_after_downsample=config.conv_after_downsample,
             use_native_resize_op=config.use_native_resize_op,
             conv_bn_relu_pattern=config.conv_bn_relu_pattern,
+            use_tpu=config.use_tpu,
```
mingxingtan, line 412: Should you update the hparams_config.py to add use_tpu?

fsx950223: No, config get ```use_tpu``` from params which get ```use_tpu``` from FLAGS.

mingxingtan: If use_tpu is not in config, other scripts like model_inspect.py or inference.py may fail (also inconvenient if people directly use efficientdet_arch rather than main).

An alternative:  change it to  "use_tpu=config.get('use_tpu', None)". WDYT?

fsx950223: My fault, I add it to hparams

mingxingtan:
Hi @fsx950223 , I was trying to use copybara to import your PR, but seems like it did not work well as expected: it did import your PR, but somehow did not keep your authorship and your commit numbers.

I am sorry for the outcome. Will investigate more about the issue. For now, I will close this PR since it is already integrated into master branch.

fsx950223:
> Hi @fsx950223 , I was trying to use copybara to import your PR, but seems like it did not work well as expected: it did import your PR, but somehow did not keep your authorship and your commit numbers.
> 
> I am sorry for the outcome. Will investigate more about the issue. For now, I will close this PR since it is already integrated into master branch.

That's nothing

HyunjiEllenPak:
@fsx950223 How did you get the training graph? 
Could you tell me the specs that you used in the training?

fsx950223:
> @fsx950223 How did you get the training graph?
> Could you tell me the specs that you used in the training?

I use noisy student efficientnet ckpt and you could follow the tutorial to get good performance.

HyunjiEllenPak:
@fsx950223 How many GPUs did you use and how did you set batch size for best performance?

fsx950223:
> @fsx950223 How many GPUs did you use and how did you set batch size for best performance?

1 GPU, batch_size=8

PR #198, opened 2020-04-10
remove useless argument

fsx950223:
```eval_after_training``` has the same behavior as ```train_and_eval```.


mingxingtan:
Hi @fsx950223 ,  eval_after_training means run one eval after all training steps, while train_and_eval means for each epoch run train+eval.

Maybe you can change the train_and_eval to make the interval epochs or steps configurable?

https://github.com/google/automl/blob/6ecb218626da627ca4c4e4bed4b760c7e66a1b70/efficientdet/main.py#L344

fsx950223:
It's a mistake

PR #201, opened 2020-04-10
Distribute training

fsx950223:
The implementation has followed bugs:
1.don't support ema
2.tensorboard bug
3.params may need to be tuned for multi GPU

### DIFF
--- a/efficientdet/dataloader.py
+++ b/efficientdet/dataloader.py
@@ -313,7 +313,7 @@ class InputReader(object):
         return (image, cls_targets, box_targets, num_positives, source_id,
                 image_scale, boxes, is_crowds, areas, classes)
 
-    batch_size = params['batch_size']
+    batch_size = params.get('batch_size', params.get('default_batch_size'))
     dataset = tf.data.Dataset.list_files(
         self._file_pattern, shuffle=self._is_training)
     if self._is_training:
--- a/efficientdet/det_model_fn.py
+++ b/efficientdet/det_model_fn.py
@@ -36,9 +36,9 @@ _DEFAULT_BATCH_SIZE = 64
 
 def update_learning_rate_schedule_parameters(params):
   """Updates params that are related to the learning rate schedule."""
-  # params.get('batch_size', params.get('default_batch_size')) is per-shard within model_fn if use_tpu=true.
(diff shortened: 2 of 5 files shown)


paulcx:
found a error "TypeError: _call_input_fn() takes 3 positional arguments but 4 were given" which might be the problem of TPUEstimator or MirroredStrategy?

fsx950223:
> An error found "TypeError: _call_input_fn() takes 3 positional arguments but 4 were given" which might be the problem of TPUEstimator or MirroredStrategy?

please install tf2.2

InfiniteLife:
I want to try this PR but not sure how. I manually added lines

```
  params['default_batch_size'] = FLAGS.train_batch_size
  if len(tf.config.experimental.list_physical_devices('GPU')) > 1:
    strategy = tf.distribute.MirroredStrategy()
  else:
    strategy = None

  run_config = tf.estimator.tpu.RunConfig(
      cluster=tpu_cluster_resolver,
      evaluation_master=FLAGS.eval_master,
      model_dir=FLAGS.model_dir,
      log_step_count_steps=FLAGS.iterations_per_loop,
      session_config=config_proto,
      tpu_config=tpu_config,
      train_distribute=strategy
  )
```
to my copy of repository and got tf 2.2.0-rc version of tensorflow, but now when I am running training I get 

```

> tensorflow.python.autograph.pyct.error_utils.KeyError: in user code:
> ....
> hparams_config.py:48 __getattr__
>         return self.__dict__[k]
> 
>     KeyError: 'is_tensor_like'

```

LucasSloan:
I checked out this branch, but I ran into some error with NCCL.  I tried switching to tf.distribute.HierarchicalCopyAllReduce(), but that also errored out with a CUDA out of memory error.  The error occurred even with the smallest feasible batch size (2, one per gpu).

I'm using 2x 2080 Ti's on the tf 2.2 rc, using the tensorflow/tensorflow:2.2.0rc3-gpu docker container.

LucasSloan:
Update:

I just tried again with the latest version of this code, and I had exactly the same issue with an OOM error.  I then tried disabling XLA, and it worked, albeit with few samples per second than with a single GPU with XLA on.

Is there anything I can do to help with the keras stuff?

PR #232, opened 2020-04-14
enable native nms as default

fsx950223:
Enable native nms as default, because it's much faster than nms_tf and they get same result.


mingxingtan:
Thanks a lot, @fsx950223

PR #240, opened 2020-04-15
Data format

### LINKED ISSUE #54: Does model support channels_first data_format?

fsx950223:
Add channel first support.
Fix #54.

### DIFF
--- a/efficientdet/inference.py
+++ b/efficientdet/inference.py
@@ -372,6 +372,8 @@ class ServingDriver(object):
         images.append(image)
       scales = tf.stack(scales)
       images = tf.stack(images)
+      if params['data_format'] == 'channels_first':
+        images = tf.transpose(images, [0, 3, 1, 2])
       class_outputs, box_outputs = build_model(self.model_name, images,
                                                **params)
       params.update(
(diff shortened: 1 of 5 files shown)


fsx950223:
It's 0.0794ms on GPU with batch_size=1, FPS: 12.594458438287154
It's 0.0462ms on GPU with batch_size=32, FPS: 21.62466738705676
GTX 1080TI, NCHW.

efficientdet/inference.py
```diff
@@ -337,7 +337,7 @@ def __init__(self,
 
     self.signitures = None
     self.sess = None
-    self.disable_pyfun = True
+    self.disable_pyfun = False
```
mingxingtan, line 340: Is it intended to set disable_pyfun as False? Isn't tf version better?

efficientdet/inference.py
```diff
@@ -520,7 +522,7 @@ def __init__(self,
       self.params.update(dict(image_size=image_size))
     if num_classes:
       self.params.update(dict(num_classes=num_classes))
-    self.disable_pyfun = False
+    self.disable_pyfun = True
```
mingxingtan, line 525: Also here.

fsx950223:
The improvement may come from the VOC dataset rather than NCHW. Since top_k's performance highly depends on num_classes.

PR #245, opened 2020-04-16
fix annotations in inference.py

Jupanlee:
Hi! I fix Example 2 in ServingDriver, in which serve_files should be replaced with serve_images.


mingxingtan:
Thanks for the fix. Could you sign the CLA so I can merge this PR? Thanks!

mingxingtan:
Oh, looks like this is the same as the other PR. Closing this one now.

PR #246, opened 2020-04-16
Fcos

fsx950223:
I try to integrate with fcos. Could you help me to finetune the model. Thanks.


fsx950223:
Haha, I have found better solution on faster rcnn. I will open source it later. Close it temporarily.

PR #266, opened 2020-04-20
fix channels bug

fsx950223:
I forget to add transpose when I submit pull request.
@mingxingtan

### DIFF
--- a/efficientdet/det_model_fn.py
+++ b/efficientdet/det_model_fn.py
@@ -122,19 +122,18 @@ def learning_rate_schedule(params, global_step):
                                 params['lr_warmup_step'],
                                 params['first_lr_drop_step'],
                                 params['second_lr_drop_step'], global_step)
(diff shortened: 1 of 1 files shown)


mingxingtan:
Thanks @fsx950223 Does it work with saved_model_infer now?

fsx950223:
> Thanks @fsx950223 Does it work with saved_model_infer now?

I just test infer. I believe they have the same behavior and you need to export new saved model.

mingxingtan:
Just tested, seems still not working on my side. Do you just change the data_format to 'channel_first' in hparams_config.py?

fsx950223:
> Just tested, seems still not working on my side. Do you just change the data_format to 'channel_first' in hparams_config.py?

yes

fsx950223:
Screenshot from 2020-04-20 16-36-24

PR #269, opened 2020-04-20
fix upsample

fsx950223:
Please check it.
https://colab.research.google.com/drive/1D51aRE8VZc3pk8Fz4i7NlKI0VHY96y7L#scrollTo=jGKs3w2_ZXnu


fsx950223:
Maybe I should add some test cases.

PR #277, opened 2020-04-21
Add a few simple tests.

mingxingtan:
PiperOrigin-RevId: 307515662


fsx950223:
> PiperOrigin-RevId: 307515662



> PiperOrigin-RevId: 307515662

I got 65fps(NHWC) on colab with batch_size=1.
I got 65fps(NCHW) on colab with batch_size=1.
Amazing performance.

PR #293, opened 2020-04-23
Id mapping

fsx950223:


mingxingtan:
Hi @fsx950223, I actually like this PR, but it would be better to make the id_mapping as a json file. WDYT?

fsx950223:
> WDYT

I just mean id mapping could be thrown out. Finally, it depends on you. BTW, JSON has to use string as key.

PR #309, opened 2020-04-25
Accept custom hyper-parameter files 

Ely-S:
When developing custom models, I had to edit the `hparams_config.py` file because I was changing too many parameters to use the `--hparams` option. The hyper-parameters also need to stay the same between training and inference, and `inspect_model.py` did not take the `--hparams` option. This made pulling updates hard because of merge conflicts in the config file.

I made `model_inspect.py` accept the `--hparams` option, and modified `--hparams` to take either a python module or key-value pairs.

Example:

```
python model_inspect.py --model_name efficientdet-d0  \
  --runmode=infer --ckpt_path=ckpt \
  --hparams=config  \
  --input_image=test.jpg  --output_image_dir=/tmp
```

Where `config.py` is a file containing the hyper-parameters for the scaled-down model I want to use to recognize persons and bicycles.

```
backbone_name='efficientnet-lite0'
image_size=256
fpn_num_filters=36
fpn_cell_repeats=2
box_class_repeats=2
image_size=256
num_classes=20

label_id_mapping={
  15: "person",
  2: "bicycle"
}
```

This approach makes keeping track of custom hyper-parameters easier and seems like it could benefit others.

I'd be happy to make any changes requested to merge this in. If this does not fit within the design goals of this project, feel free to just close this PR.

### DIFF
--- a/efficientdet/hparams_config.py
+++ b/efficientdet/hparams_config.py
@@ -47,6 +47,9 @@ class Config(object):
   def __getattr__(self, k):
     return self.__dict__[k]
 
+  def __getitem__(self, k):
+    return self.__dict__[k]
+
   def __repr__(self):
     return repr(self.as_dict())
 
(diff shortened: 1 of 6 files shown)


efficientdet/hparams_config.py
```diff
@@ -132,6 +157,7 @@ def default_detection_configs():
 
   # input preprocessing parameters
   h.image_size = 640   # An integer or a string WxH such as 640x320.
+  h.batch_size = 1
   h.input_rand_hflip = True
```
mingxingtan, line 161: batch_size is reserved for TPU training, so we cannot add batch_size here.

efficientdet/hparams_config.py
```diff
@@ -140,6 +166,7 @@ def default_detection_configs():
   # dataset specific parameters
   h.num_classes = 90
   h.skip_crowd_during_training = True
+  h.label_id_mapping = None
```
mingxingtan, line 169: How about make label_id_mapping as a json file name? (seems not easy to pass a huge dict to hparams).

efficientdet/model_inspect.py
```diff
@@ -40,7 +40,9 @@
 flags.DEFINE_string('logdir', '/tmp/deff/', 'log directory.')
 flags.DEFINE_string('runmode', 'dry', 'Run mode: {freeze, bm, dry}')
 flags.DEFINE_string('trace_filename', None, 'Trace file name.')
-flags.DEFINE_integer('num_classes', 90, 'Number of classes.')
+flags.DEFINE_integer('num_classes', None,
+                     'Number of classes. Defaults to 90 '
```
mingxingtan, line 44: How about deleting num_classes, input_image_size, enable_ema, and data_format (they can simply override with hparams)? WDYT?

efficientdet/hparams_config.py
Ely-S: I like that a lot

efficientdet/model_inspect.py
Ely-S: That would be great. It would be simpler to have just one source of configuration.

Ely-S:
Good suggestions @mingxingtan. I made the changes and updated the README with an example.

mingxingtan:
Hi @Ely-S Thanks for the PR.

I am going to merge this PR, but just for a heads up, after the merging, I will immediately send out another PR what will do some cleanup and updates.

PR #313, opened 2020-04-26
add freeze training

### LINKED ISSUE #273: Freeze training
Maybe there should be an option to freeze backbone or backbone+fpn for fine-tuning model

fsx950223:
I'm not sure whether I should freeze ema?
Fix #273.

### DIFF
--- a/efficientdet/efficientdet_arch.py
+++ b/efficientdet/efficientdet_arch.py
@@ -25,6 +25,7 @@ from __future__ import print_function
 
 import functools
 import itertools
+import re
 from absl import logging
 import numpy as np
 import tensorflow.compat.v1 as tf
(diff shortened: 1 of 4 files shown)


efficientdet/hparams_config.py
```diff
@@ -200,6 +200,8 @@ def default_detection_configs():
 
   h.backbone_name = 'efficientnet-b1'
   h.backbone_config = None
+  # Availabel values are 'fpn', 'backbone', None
+  h.freeze = None
```
mingxingtan, line 204: How about provide something like  "h.var_freeze_expr" (we can provide some examples in comments and tests)?

Then you can just define one function like:

def variable_filter_fn(variables, var_freeze_expr=None):
  matcher = re.compile(var_freeze_expr) if var_freeze_expr else None
  var_list = [v for v in variables if matcher and not matcher.match(v.name)]
  return var_list

efficientdet/efficientdet_arch.py
```diff
@@ -694,3 +694,34 @@ def efficientdet(features, model_name=None, config=None, **kwargs):
       *utils.num_params_flops()))
 
   return class_outputs, box_outputs
+
+
+def freeze_backbone(variables):
```
mingxingtan, line 699: It seems a little over complicated to define multiple functions, how about allowing user to define regexpr as I suggested in hparams.py?

fsx950223:
> Thank you, @fsx950223

I change them later.

PR #324, opened 2020-04-28
add dynampic batch size inference

fsx950223:
```tf.image.combined_non_max_suppression```
cc @mingxingtan @itsliupeng

### DIFF
--- a/efficientdet/inference.py
+++ b/efficientdet/inference.py
@@ -77,6 +77,14 @@ def image_preprocess(image, image_size: Union[int, Tuple[int, int]]):
   image_scale = input_processor.image_scale_to_original
   return image, image_scale
 
+@tf.autograph.to_graph
+def batch_image_files_decode(image_files):
+  raw_images = tf.TensorArray(tf.uint8, size=0, dynamic_size=True)
+  for i in tf.range(tf.shape(image_files)[0]):
+    image = tf.io.decode_image(image_files[i])
+    image.set_shape([None, None, None])
+    raw_images = raw_images.write(i, image)
+  return raw_images.stack()
 
 def batch_image_preprocess(raw_images,
                            image_size: Union[int, Tuple[int, int]],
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: efficientdet/inference.py [hunk 1]
```diff
@@ -213,6 +213,55 @@ def det_post_process(params: Dict[Any, Any],
...
+    boxes = tf.expand_dims(boxes, axis=2)
+    nmsed_boxes, nmsed_scores, nmsed_classes, valid_detections = tf.image.combined_non_max_suppression(
+      boxes, scores,
+      anchors.MAX_DETECTIONS_PER_IMAGE,
+      anchors.MAX_DETECTIONS_PER_IMAGE,
+      score_threshold=min_score_thresh,
+      clip_boxes=False)
+    height = nmsed_boxes[..., 2] - nmsed_boxes[..., 0]
+    width = nmsed_boxes[..., 3] - nmsed_boxes[..., 1]
+    detections = tf.stack([
```
mingxingtan, line 253: Why do we need a tf.tile here?  Could you add a simple comment?

mingxingtan: Actually, the whole image_id part can be just " image_ids = tf.ones([batch_size, max_boxes_to_draw], dtype=tf.float32)", since the image_ids are fake anyway for inference.

mingxingtan:
Hi @fsx950223 I tried to test your PR, but it seems if batch size is None in the saved model, then the output image is not as expected. I tried to sync to the latest code and add a few minor fixes, but still failed to get the right image output.

Could you help test the PR? Thanks!

fsx950223:
> Hi @fsx950223 I tried to test your PR, but it seems if batch size is None in the saved model, then the output image is not as expected. I tried to sync to the latest code and add a few minor fixes, but still failed to get the right image output.
> 
> Could you help test the PR? Thanks!

Yes, I test it later.

fsx950223:
> Hi @fsx950223 I tried to test your PR, but it seems if batch size is None in the saved model, then the output image is not as expected. I tried to sync to the latest code and add a few minor fixes, but still failed to get the right image output.
> 
> Could you help test the PR? Thanks!

I got SOTA result with following command:
```
python model_inspect.py --runmode=infer --model_name=efficientdet-d0 --ckpt_path=/media/fangsixie/data/download/efficientdet-d0 --input_image=./test.png --output_image_dir=./testdata --hparams="image_size=1920x1280" --max_boxes_to_draw=100 --min_score_thresh=0.2 
```

efficientdet/inference.py
fsx950223: I'm not sure whether developer will use image_id. If use ```tf.ones``` they could only get 1.

efficientdet/inference.py
```diff
@@ -751,7 +751,7 @@ def inference(self, image_path_pattern: Text, output_dir: Text, **kwargs):
 
       # for postprocessing.
       params.update(
-          dict(batch_size=len(raw_images), disable_pyfun=self.disable_pyfun))
+          dict(batch_size=None, disable_pyfun=self.disable_pyfun))
```
fsx950223, line 754: Test batch nms

fsx950223:
I have fix the saved model inference.

mingxingtan:
Excellent!  Thank you, @fsx950223 !

I noticed that there are some very minor differences, but nothing major.  I am going to merge this PR now, and immediately follow up with a commit for a few minor format changes.

PR #329, opened 2020-04-29
Clean up and some followup tests for config-based inference.

mingxingtan:
Mainly follow up https://github.com/google/automl/pull/309

### DIFF
--- a/efficientdet/g3doc/coco_ids.yaml
+++ b/efficientdet/g3doc/coco_ids.yaml
@@ -1,0 +1,81 @@
+0: background
+1: person
+2: bicycle
+3: car
+4: motorcycle
+5: airplane
+6: bus
+7: train
+8: truck
+9: boat
+10: traffic light
+11: fire hydrant
+13: stop sign
+14: parking meter
+15: bench
+16: bird
+17: cat
+18: dog
+19: horse
+20: sheep
+21: cow
+22: elephant
+23: bear
+24: zebra
+25: giraffe
+27: backpack
+28: umbrella
+31: handbag
+32: tie
+33: suitcase
+34: frisbee
+35: skis
+36: snowboard
+37: sports ball
+38: kite
+39: baseball bat
+40: baseball glove
+41: skateboard
+42: surfboard
+43: tennis racket
+44: bottle
+46: wine glass
+47: cup
+48: fork
+49: knife
+50: spoon
+51: bowl
+52: banana
+53: apple
+54: sandwich
+55: orange
+56: broccoli
+57: carrot
+58: hot dog
+59: pizza
+60: donut
+61: cake
+62: chair
+63: couch
+64: potted plant
+65: bed
+67: dining table
+70: toilet
+72: tv
+73: laptop
+74: mouse
+75: remote
+76: keyboard
+77: cell phone
+78: microwave
+79: oven
+80: toaster
+81: sink
+82: refrigerator
+84: book
+85: clock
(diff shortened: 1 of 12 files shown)


mingxingtan:
@Ely-S Please take a look this PR.
@fsx950223 This PR allows hparams to accept yaml file, where keys can be integer as well.

fsx950223:
LGTM

fsx950223:
Don't use JSON and YAML together

mingxingtan:
Did not find a good way to dump to a string with yaml.

mingxingtan:
NM, yaml support that, just need to change dumps to dump.

mingxingtan:
Thanks for the quick review @fsx950223 

@Ely-S I am going to merge this PR, but feel free to add your comments, and we can update the code accordingly later.

efficientdet/main.py
```diff
@@ -141,13 +141,11 @@ def main(argv):
...
-  if FLAGS.num_epochs:  # TODO: remove this flag after updating all docs.
+  if FLAGS.num_epochs:  # NOTE: remove this flag after updating all docs.
     config.num_epochs = FLAGS.num_epochs
 
-  if isinstance(config.image_size, str) and 'x' in config.image_size:
-    # image size is in format of width x height.
-    width, height = config.image_size.split('x')
-    config.image_size = (int(height), int(width))
+  # Parse image size in case it is in string format.
+  config.image_size = utils.parse_image_size(config.image_size)
```
Ely-S, line 148: nice refactor

efficientdet/g3doc/coco_ids.yaml
```diff
@@ -0,0 +1,81 @@
+0: background
```
Ely-S, line 1: This is much nicer than a JSON file since keys can be ints.

Ely-S:
@mingxingtan This is great.

I see that `tf.io.GFile` is the preferred way to open files. I'll use that next time.

I'm going test this by running some training/inference typical at my day job.

efficientdet/hparams_config.py
```diff
@@ -91,32 +94,45 @@ def override(self, config_dict_or_str):
...
+      if not attr.startswith('_'):
         config_dict[attr] = getattr(module, attr)
 
     return config_dict
 
-  def parse_from_str(self, config_str):
+  def parse_from_yaml(self, yaml_file_path: Text) -> Dict[Any, Any]:
+    """Parses a yaml file and returns a dictionary."""
+    with tf.io.gfile.GFile(yaml_file_path, 'r') as f:
+      config_dict = yaml.load(f, Loader=yaml.FullLoader)
```
Ely-S, line 127: If this library is pyyaml, `yaml.safe_load` is generally preferred.

Ely-S:
@mingxingtan It seems like yaml is not part of the python standard library, so this change is breaking on my system.

```
  File "/home/eli/ondaka/edet/automl/efficientdet/hparams_config.py", line 28, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
```

How would you feel about adding a requirements.txt file?

efficientdet/hparams_config.py
mingxingtan: You are right. safe_load seems to be better.

mingxingtan:
@Ely-S Good idea, could you help preparer a requirements.txt file?

Ely-S:
sure thing

PR #335, opened 2020-04-29
add requirements.txt

Ely-S:
Follow-up to  #329


mingxingtan:
Thanks @Ely-S!  Do we really need so many packages? Can we simplify it to minimal (say if pip install tensorflow will automatically install some other packages, then we don't need to list them).

Maybe just:   tensorflow, pycocotools, yaml,  or other major packages?

Ely-S:
That requirements.txt contains dependencies of dependencies. Here's a new version that has just the packages which are imported directly. The trade-off for a cleaner requirements.txt is that builds become less deterministic.

I'm conflicted about including tensorflow in the requirements.txt so I removed it. The user may need a different version of TF than in the requirements file. Here are some cases in which that may true:

1. The code needs to run on a ctpu instance. `ctpu up` creates a compute instance and a TPU cluster of the same TF version. The compute instance created by `ctpu up` comes with TF pre-installed and it's best not to modify it. I've encountered errors by installing a custom TF version on the compute instance, because the version of TF on the compute instance and TPU cluster need to be the same. Pinning a TF version in requirements.txt might break most cptu installations.
2. The user may require tensorflow-gpu instead of tensorflow
3. The user may require TF1, and it seems like the codebase currently supports both versions
4. If tensorflow-addons could be included, but only for TF2

@mingxingtan what do you think?

mingxingtan:
Looks much better. Merging this PR now.
I may just add the latest tensorflow later.

PR #362, opened 2020-05-06
add augmix

fsx950223:
https://arxiv.org/pdf/1912.02781.pdf
https://github.com/google-research/augmix.

### DIFF
--- a/efficientdet/aug/autoaugment_test.py
+++ b/efficientdet/aug/autoaugment_test.py
@@ -26,7 +26,6 @@ from aug import autoaugment
 class AutoaugmentTest(tf.test.TestCase):
 
   def test_autoaugment_policy(self):
-    tf.disable_eager_execution()
     # A very simple test to verify no syntax error.
     image = tf.placeholder(tf.uint8, shape=[640, 640, 3])
     bboxes = tf.placeholder(tf.float32, shape=[4, 4])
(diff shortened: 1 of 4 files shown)


efficientdet/aug/autoaugment_test.py
```diff
@@ -26,10 +26,12 @@
 class AutoaugmentTest(tf.test.TestCase):
 
   def test_autoaugment_policy(self):
+    tf.disable_eager_execution()
```
mingxingtan, line 29: Since tf.disable_eager_execution() has already run before tf.test.main(), you probably don't need tf.disable_eager_execution() here?

efficientdet/hparams_config.py
```diff
@@ -177,6 +177,10 @@ def default_detection_configs():
   h.train_scale_min = 0.1
   h.train_scale_max = 2.0
   h.autoaugment_policy = None
+  h.use_augmix = False
```
mingxingtan, line 180: Do people need to change mixture_width/mixture_depth/alpha?
If people often need to change it, then having 3 options are better, but if most time they don't need to change them, then it might be better to just use a single option:  h.augmix_params = (1, 3, -1)   # (alpha, width, depth)
WDYT?

mingxingtan:
Very nice. Thank you for the PR @fsx950223

PR #376, opened 2020-05-07
Change some utils to tf.keras.layers

VicariNishimura:
This is essentially a test (first PR), and those functions in utils need to be class that inherit from tf.keras.layers.Layer before changing BoxNet/ClassNet. 

@mingxingtan @Ely-S @fsx950223 if anything wrong tell me (want to help not to take more of your time)

PR #323, opened 2020-04-28
keras refact

fsx950223:
Refact with keras and get same output with tf1.x.

PR #381, opened 2020-05-07
Wrap build_bifpn_layer in keras layer

Ely-S:
This shares the intention of #323  and #376  in that it seeks to create a keras implementation. This code should be enough to use a BiFPN in a keras model.

I started implementing `build_bifpn_layer` as pure keras code, but the task started to get too big. It might make sense to refactor it into smaller pieces first. Here is what I have working so far, hoping it's a start.

@fsx950223 @mingxingtan Any suggestions on getting it to the next level?


mingxingtan:
Thank you, @Ely-S and @fsx950223. I agree that we should check in small PRs step by step.

We have two potential ways: (1) create a branch with name "keras" and all changes check to this branch; (2) create a folder name "keras" in master branch, then all these CLs check in to that folder. I personally prefer approach 2 for its simplicity. WDYT?

mingxingtan:
I created a subfolder "keras", could you help move your code to that folder? Thanks! @Ely-S

Ely-S:
@mingxingtan I also like the idea of a submodule because it will allow us to import the latest code from the existing implementation. I'll move this into a keras folder.

Ely-S:
done

mingxingtan:
Thanks for the PR, @Ely-S

efficientdet/efficientdet_arch.py
```diff
@@ -591,6 +591,43 @@ def get_fpn_config(fpn_name, min_level, max_level, weight_method):
...
+    nodes: a list of tensorflow features at different levels
+    weight_method: feature fusion method. One of:
+      * "attn" - Softmax weighted fusion
+      * "fastattn" - Fast normalzied feature fusion
+      * "sum" - a sum of inputs
+  """
+  dtype = nodes[0].dtype
+
+  if weight_method == 'attn':
+    edge_weights = [tf.cast(tf.Variable(1.0, name='WSM'), dtype=dtype)
```
fsx950223, line 608: Variable shouldn't be created in ```call```.

PR #390, opened 2020-05-09
Adam

fsx950223:
I found Adam could get better performance on custom dataset.

### DIFF
--- a/efficientdet/anchors.py
+++ b/efficientdet/anchors.py
@@ -456,7 +456,7 @@ class Anchors(object):
     self.num_scales = num_scales
     self.aspect_ratios = aspect_ratios
     self.anchor_scale = anchor_scale
-    image_size = utils.parse_image_size(image_size)
+    self.image_size = utils.parse_image_size(image_size)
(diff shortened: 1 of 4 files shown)


efficientdet/anchors.py
```diff
@@ -335,6 +335,10 @@ def _generate_detections_tf(cls_outputs,
     scores = detections[:, 4]
     boxes = detections[:, :4]
 
+  if isinstance(image_size, int):
```
mingxingtan, line 338: You can replace this with "image_size = utils.parse_image_size(image_size)"

efficientdet/inference.py
```diff
@@ -249,6 +249,10 @@ def det_post_process_combined(params, cls_outputs, box_outputs, scales,
           tf.expand_dims(tf.range(batch_size), axis=1), [1, max_boxes_to_draw]),
       dtype=tf.float32)
   image_size = params['image_size']
+  if isinstance(image_size, int):
```
mingxingtan, line 252: image_size = utils.parse_image_size(image_size) is simpler.

PR #397, opened 2020-05-12
Change BoxNet/ClassNet to Keras

VicariNishimura:
@fsx950223 

this PR is a proper version of the old one. (updated with last commits).


fsx950223:
Could you review the PR.@LucasSloan

efficientdet/keras/utils_keras.py
```diff
@@ -0,0 +1,129 @@
...
+
+import tensorflow.compat.v1 as tf
+
+# pylint: disable=logging-format-interpolation
+from utils import BatchNormalization, TpuBatchNormalization
+
+
+class ActivationFn(tf.keras.layers.Layer):
+  def __init__(self, act_type: Text, name='activation_fn', **kwargs):
+
```
fsx950223, line 30: Add utils test cases too.

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -81,7 +88,7 @@ def get_config(self):
     }
 
 
-class ResampleFeatureMap(tf.keras.layers.Layer):
```
mingxingtan, line 84: Are tfv1.keras.layers the same as tf.keras.layers? If they are the same, you can just use tf.keras.layers.

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -210,3 +217,410 @@ def get_config(self):
...
+               min_level=3,
+               max_level=7,
+               is_training=False,
+               act_type='swish',
+               repeats=4,
+               separable_conv=True,
+               survival_prob=None,
+               use_tpu=False,
+               data_format='channels_last',
+               name='class_net', **kwargs):
```
mingxingtan, line 237: Would be great if you can add some doc string.

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -210,3 +217,410 @@ def get_config(self):
...
+          kernel_initializer=tf.random_normal_initializer(stddev=0.01),
+          data_format=self.data_format,
+          kernel_size=3,
+          activation=None,
+          bias_initializer=tf.constant_initializer(
+              -np.math.log((1 - 0.01) / 0.01)),
+          padding='same',
+          name=f'{self.name}/class-predict')
+
+  def call(self, inputs, **kwargs):
```
mingxingtan, line 326: Hmm, our comment docstring format is like this:

def function(x, y):
  """One-line description of the function.

  Args:
     x: what is x
     y: what is y

  Returns:
     xxx
"

efficientdet/keras/efficientdet_arch_keras.py
VicariNishimura: No, there are some diffrences, not much but for example this layer will not work with tf.keras.layers.Layer as the call is not call(self, inputs, **kwargs).

fsx950223: It seems tf1 Layer and tf2 Layer are same class?

VicariNishimura: All i'm sure is that if I change tf.compat.v1 by tf I got problem with matching signature of base method in class Layer

fsx950223: Yeah, you could change all tf.compat.v1 to tf if necessary.

efficientdet/keras/utils_keras.py
```diff
@@ -0,0 +1,129 @@
...
+    elif act_type == 'swish_native':
+      self.act = lambda x: x * tf.sigmoid(x)
+    elif act_type == 'relu':
+      self.act = tf.nn.relu
+    elif act_type == 'relu6':
+      self.act = tf.nn.relu6
+    else:
+      raise ValueError('Unsupported act_type {}'.format(act_type))
+
+    self.layer = tf.keras.layers.Lambda(lambda x: self.act(x), name=name)
```
fsx950223, line 46: Remove Lambda.
Over wrapper make developers annoyed.

VicariNishimura: My bad it was just for the naming tests, it have no purpose, thx for pointing it out

efficientdet/keras/utils_keras.py
```diff
@@ -0,0 +1,129 @@
...
+                                      epsilon=epsilon,
+                                      center=True,
+                                      scale=True,
+                                      gamma_initializer=self.gamma_initializer,
+                                      name=f'{parent_name}/{name}')
+
+    self.act = ActivationFn(act_type, name=parent_name)
+
+  def call(self, inputs, **kwargs):
+    x = self.layer.apply(inputs, training=self.training)
```
fsx950223, line 107: use __call__, instead of ```apply``` and ```call```.

VicariNishimura: Using it will change the names

fsx950223: Just keep variable names same as origin one.
I believe it won't influence it.

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -210,3 +217,468 @@ def get_config(self):
...
+  act_type: String of the activation used
+  data_format: channel_first if [B, C, W, H] format, 'channels_last' if [B, W, H, C]
+  box_class_repeats: number of "intermediate" layers
+  separable_conv: True to use separable_conv instead of conv2D
+  survival_prob: if a value is set then drop connect will be used with this probability
+  use_tpu: True to use TPU compatible functions
+
+  """
+  def __init__(self,
+               num_classes=80,
```
LucasSloan, line 241: num_classes = 80?

Isn't the default in this codebase 90?

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -210,3 +217,468 @@ def get_config(self):
...
+            init_zero=False,
+            use_tpu=self.use_tpu,
+            data_format=self.data_format,
+            name='class-%d-bn-%d' % (i, level))
+      self.bn_act_ops.append(bn_act_ops_per_level)
+
+    if self.use_dc:
+      self.dc = keras.utils_keras.DropConnect(self.survival_prob)
+
+    if self.separable_conv:
```
LucasSloan, line 315: Can you decide between regular and separable_conv once, instead of having 2 of these if blocks?

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -210,3 +217,468 @@ def get_config(self):
...
+            filters=self.num_filters,
+            kernel_initializer=tf.random_normal_initializer(stddev=0.01),
+            data_format=self.data_format,
+            kernel_size=3,
+            activation=None,
+            bias_initializer=tf.zeros_initializer(),
+            padding='same',
+            name=f'class-%d' % i))
+
+      # Level only apply here so it's maybe better inside (no need to use tf.AUTO_REUSE anymore)
```
LucasSloan, line 300: I don't understand this comment - could you explain further?

efficientdet/keras/efficientdet_arch_keras.py
VicariNishimura: I left it for retro-compatibility issues

efficientdet/keras/efficientdet_arch_keras.py
VicariNishimura: I can remove it, it was more of a self comment regarding the fact that in the original code level is used outside of the box/class, but now it is inside.

efficientdet/keras/efficientdet_arch_keras.py
VicariNishimura: Yes, i'm just used to 80, but as 90 is the default config one it shouldn't have any influence, I'll change it and it'll come within next commit

mingxingtan: Hi please use 90 classes to avoid confusion. COCO 2012 is not simply using class 0-80, see this: https://tech.amikelive.com/node-718/what-object-categories-labels-are-in-coco-dataset/. Thank you!

mingxingtan:
@VicariNishimura @fsx950223 @LucasSloan @Ely-S 
Is this PR ready to merge?

mingxingtan:
@VicariNishimura @fsx950223 
Since keras layers are in progress and this PR doesn't change the main code path, I would like to merge this PR for now to unblock other possible changes.  We can still update it later if needed. Thanks!

PR #408, opened 2020-05-14
Add get_model to backbone builders

VicariNishimura:
This small change allow to get the backbone model to use it later instead of a function needing features.


efficientdet/backbone/efficientnet_builder.py
```diff
@@ -325,3 +325,54 @@ def build_model_base(images, model_name, training, override_params=None):
   model = efficientnet_model.Model(blocks_args, global_params, model_name)
   features = model(images, training=training, features_only=True)
   return features, model.endpoints
+
+
+def get_model(model_name,
```
mingxingtan, line 330: How about move this get_model to "https://github.com/google/automl/blob/master/efficientdet/backbone/backbone_factory.py"? Then you don't need to define this in each builder file?

efficientdet/backbone/efficientnet_builder.py
```diff
@@ -325,3 +325,54 @@ def build_model_base(images, model_name, training, override_params=None):
...
+  if not training or fine_tuning:
+    if not override_params:
+      override_params = {}
+    override_params['batch_norm'] = utils.BatchNormalization
+    if fine_tuning:
+      override_params['relu_fn'] = functools.partial(swish, use_native=False)
+  blocks_args, global_params = get_model_params(model_name, override_params)
+
+  if model_dir:
+    param_file = os.path.join(model_dir, 'model_params.txt')
```
mingxingtan, line 366: Maybe use tf.io.gfile.exists?

efficientdet/backbone/efficientnet_builder.py
```diff
@@ -325,3 +325,54 @@ def build_model_base(images, model_name, training, override_params=None):
...
+    if not tf.gfile.Exists(param_file):
+      if not tf.gfile.Exists(model_dir):
```
mingxingtan, line 368: tf.io.gfile.mkdirs

PR #416, opened 2020-05-18
Change the 'MAX_NUM_INSTANCES=100' with user input data maximum instances or add a flag to get user input for the same

Samjith888:
Got the following error 
**tensorflow.python.framework.errors_impl.InvalidArgumentError: assertion failed: [533]**

The error is in the following error ''MAX_NUM_INSTANCES=100" 
https://github.com/google/automl/blob/master/efficientdet/dataloader.py#L27

we can solve this error by changing maximum number of instances based on our custom dataset. Please automate this


mingxingtan:
Hi @Samjith888 I did not fully understand this PR. Looks like it contains a lot of unrelated changes. Could you create a PR from the head (of master)? Thanks !

PR #413, opened 2020-05-18
add horovod

fsx950223:
Awesome distribute strategy.
Fix #400,#28,#274,#261.

PR #421, opened 2020-05-19
Use flag to control use_horovod.

mingxingtan:
Followup to https://github.com/google/automl/pull/413

### DIFF
--- a/efficientdet/main.py
+++ b/efficientdet/main.py
@@ -24,18 +24,7 @@ from absl import app
 from absl import flags
 from absl import logging
 
-try:
-  import horovod.tensorflow as hvd
-
-  logging.info("Use horovod with multi gpus")
-  hvd.init()
-  os.environ['CUDA_VISIBLE_DEVICES'] = str(hvd.local_rank())
-  use_horovod = True
-except:
-  use_horovod = False
-
 import numpy as np
-import tensorflow.compat.v1 as tf
 
 import dataloader
 import det_model_fn
@@ -119,6 +108,7 @@ flags.DEFINE_bool('eval_after_training', False, 'Run one eval after the '
 flags.DEFINE_integer(
     'tf_random_seed', None, 'Sets the TF graph seed for deterministic execution'
     ' across runs (for debugging).')
+flags.DEFINE_bool('use_horovod', None, 'Use horovod for multi-gpu training')
 
 # For Eval mode
 flags.DEFINE_integer('min_eval_interval', 180,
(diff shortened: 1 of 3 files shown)


efficientdet/main.py
```diff
@@ -131,6 +121,15 @@
 
 
 def main(_):
+
+  if FLAGS.use_horovod:
```
mingxingtan, line 125: @fsx950223 use a flag to control whether to use horovod.

fsx950223:
According to paper, batch norm should replace by sync batch norm.
```python
class SyncBatchNormalization(tf.keras.layers.BatchNormalization):
  """Cross replica batch normalization."""

  def __init__(self, fused=False, **kwargs):
    if fused in (True, None):
      raise ValueError('SyncBatchNormalization does not support fused=True.')
    if not kwargs.get('name', None):
      kwargs['name'] = 'tpu_batch_normalization'
    super(SyncBatchNormalization, self).__init__(fused=fused, **kwargs)

  def _moments(self, inputs, reduction_axes, keep_dims):
    """Compute the mean and variance: it overrides the original _moments."""
    import horovod as hvd
    shard_mean, shard_variance = super(SyncBatchNormalization, self)._moments(
      inputs, reduction_axes, keep_dims=keep_dims)

    num_shards = hvd.size()
    if num_shards > 1:
      # Compute variance using: Var[X]= E[X^2] - E[X]^2.
      shard_square_of_mean = tf.math.square(shard_mean)
      shard_mean_of_square = shard_variance + shard_square_of_mean
      group_mean = hvd.allreduce(shard_mean)
      group_mean_of_square = hvd.allreduce(shard_mean_of_square)
      group_variance = group_mean_of_square - tf.math.square(group_mean)
      return (group_mean, group_variance)
    else:
      return (shard_mean, shard_variance)

  def call(self, *args, **kwargs):
    outputs = super(SyncBatchNormalization, self).call(*args, **kwargs)
    # A temporary hack for tf1 compatibility with keras batch norm.
    for u in self.updates:
      tf.add_to_collection(tf.GraphKeys.UPDATE_OPS, u)
    return outputs
```
It needs to add ```use_horovod``` to every batch norm function. Any better solution?

mingxingtan:
Good point. Maybe you can pass use_horovod around (similar to use_tpu)?  (Or you can also unify them to be a common flag and pass around the common flag.

Does tf.keras.layers.experimental.SyncBatchNormalization supporrt horovod?

fsx950223:
> Good point. Maybe you can pass use_horovod around (similar to use_tpu)? (Or you can also unify them to be a common flag and pass around the common flag.
> 
> Does tf.keras.layers.experimental.SyncBatchNormalization supporrt horovod?

```tf.keras.layers.experimental.SyncBatchNormalization``` doesn't horovod and it's a experimental layer only supports latest TensorFlow.

mingxingtan:
@fsx950223 Your horovod sync batch norm looks good to me. Do you plan to send a PR, so I can merge this to master branch?

fsx950223:
> @fsx950223 Your horovod sync batch norm looks good to me. Do you plan to send a PR, so I can merge this to master branch?

Yes, I do it later.

PR #431, opened 2020-05-21
Keras

fsx950223:
Follow fix.


mingxingtan:
Great! Thanks @fsx950223.   LGTM.

@VicariNishimura You may want to take a look as well?

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -534,107 +527,44 @@ def get_config(self):
...
-        'is_training_bn': self.is_training_bn,
-        'act_type': self.act_type,
-        'box_class_repeats': self.box_class_repeats,
-        'separable_conv': self.separable_conv,
-        'survival_prob': self.survival_prob,
-        'use_tpu': self.use_tpu,
-        'data_format': self.data_format,
-        **self.args
-    }
+  return class_outputs, box_outputs
```
VicariNishimura, line 567: I'm not confident with this change, the goal here with keras is to create a tf.keras.model wich can use all features from tf and keras, and be heavily customizable.

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -674,7 +604,7 @@ def efficientdet(features, model_name=None, config=None, **kwargs):
 
   # build class and box predictions.
   class_box = BuildClassAndBoxOutputs(**config)
-  class_outputs, box_outputs = class_box.call(fpn_feats)
+  class_outputs, box_outputs = class_box(fpn_feats)
```
VicariNishimura, line 607: you deleted BuildClassAndBoxOutputs

efficientdet/keras/efficientdet_arch_keras.py
fsx950223: You have break variable names and these layers are useless.

VicariNishimura: All weights was good so this was not critical as it was still possible to load model

VicariNishimura: Maybe you could just do your changes inside the layer ?

Ely-S:
nicely done @fsx950223 ! I love seeing PRs with more code deleted than added.

mingxingtan:
Since keras layers are wip, I am going to merge this PR in for now. But we can still iterate on different design choices later.

PR #435, opened 2020-05-21
Add sync batch norm

fsx950223:


efficientdet/main.py
```diff
@@ -52,11 +52,12 @@
     'eval_master', default='',
     help='GRPC URL of the eval master. Set to an appropriate value when running'
     ' on CPU/GPU')
-flags.DEFINE_bool('use_tpu', True, 'Use TPUs rather than CPUs/GPUs')
+flags.DEFINE_string('strategy', 'tpu', 'Use TPUs rather than CPUs/GPUs')
 flags.DEFINE_bool('use_fake_data', False, 'Use fake input.')
 flags.DEFINE_bool(
     'use_xla', False,
-    'Use XLA even if use_tpu is false.  If use_tpu is true, we always use XLA, '
+    'Use XLA even if strategy is not tpu. '
```
mingxingtan, line 59: Maybe you can merge these two lines into one line?

fsx950223:
> Excellent!

I got bad performance when I use sync batch norm. It's slower than single gpu when I try to train it with 2 gpus

fsx950223:
What's the recommended batch size?

mingxingtan:
batch size bigger is better, but I found 32 is big enough to get the best performance.  I guess as long as it is >=8, should be fine.

Sad to see sync batch norm is so slow.

fsx950223:
> batch size bigger is better, but I found 32 is big enough to get the best performance. I guess as long as it is >=8, should be fine.
> 
> Sad to see sync batch norm is so slow.

Should add an option to disable sync batch norm.

efficientdet/main.py
```diff
@@ -244,7 +243,7 @@ def _can_partition(spatial_dim):
       per_host_input_for_training=tf.estimator.tpu.InputPipelineConfig
       .PER_HOST_V2)
 
-  if FLAGS.use_horovod:
+  if FLAGS.strategy == 'gpus':
```
mingxingtan, line 246: Change all these "gpus" to "horovod"?

efficientdet/main.py
```diff
@@ -52,11 +52,12 @@
     'eval_master', default='',
     help='GRPC URL of the eval master. Set to an appropriate value when running'
     ' on CPU/GPU')
-flags.DEFINE_bool('use_tpu', True, 'Use TPUs rather than CPUs/GPUs')
+flags.DEFINE_string('strategy', 'tpu', 'Use TPUs rather than CPUs/GPUs')
```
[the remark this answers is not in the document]
mingxingtan: Sorry, you should better use enum:

```
flags.DEFINE_enum('strategy', 'tpu', ['tpu', 'horovod', None],
                  'Training: horovod for multi-gpu, if None, use TF default.')
```

mingxingtan:
Could you also help update all files (replace use_tpu with strategy) under keras/ ? Thanks!

mingxingtan:
Thanks @fsx950223! Could update the tests, I tried to run the tests and some tests (like keras/utils_test.py) seem to be failing.

mingxingtan:
@fsx950223 
There are still a few minor issues about strategy, but I am going to merge this PR now, but immediately follow up with another PR.

PR #439, opened 2020-05-22
fix inference_time calculation bug

jonbakerfish:
fix inference_time calculation bug


jonbakerfish:
@googlebot I signed it!

mingxingtan:
Thanks!

PR #444, opened 2020-05-23
Cleanup the strategy and sync batch norm.

mingxingtan:
A follow up to https://github.com/google/automl/pull/435

### DIFF
--- a/efficientdet/keras/utils_keras.py
+++ b/efficientdet/keras/utils_keras.py
@@ -1,25 +1,44 @@
+# Lint as: python3
+# Copyright 2020 Google Research. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Common keras utils."""
+# gtype import
+
 from typing import Text
 import tensorflow as tf
+import utils
 
 
-def batch_normalization(is_training_bn: bool,
-                        init_zero: bool = False,
-                        data_format: Text = 'channels_last',
-                        momentum: float = 0.99,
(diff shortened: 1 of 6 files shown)


efficientdet/keras/utils_keras_test.py
```diff
@@ -13,27 +13,28 @@
...
-  )
-  def test_batch_normalization(self, is_training_bn):
+  @parameterized.named_parameters(
+      ('train_local', True, ''), ('eval_local', False, ''),
+      ('train_tpu', True, 'tpu'), ('eval_tpu', False, 'tpu'))
+  def test_batch_norm(self, is_training, strategy):
     inputs = tf.random.uniform([8, 40, 40, 3])
-    with self.subTest(is_training_bn=is_training_bn):
-      expect_results = utils.batch_norm_act(inputs, is_training_bn, None)
-      tf.keras.backend.set_learning_phase(int(is_training_bn))
```
mingxingtan, line 33: @fsx950223 I believe set_learning_phase is going to be deprecated, so I changed it to directly call bn with is_training.

efficientdet/keras/utils_keras.py
```diff
@@ -30,15 +49,9 @@ def batch_normalization(is_training_bn: bool,
   else:
     gamma_initializer = tf.ones_initializer()
 
-  if data_format == 'channels_first':
-    axis = 1
-  else:
-    axis = -1
-  if strategy in ['tpu', 'gpu'] and is_training_bn:
-    batch_norm_class = tf.keras.layers.experimental.SyncBatchNormalization
```
mingxingtan, line 38: I haven't verified tf.keras.layers.experimental.SyncBatchNormalization can work well for TPU or horovod, so maybe we can just use the existing bn for now?

efficientdet/keras/utils_keras.py
```diff
@@ -1,25 +1,44 @@
...
 
-def batch_normalization(is_training_bn: bool,
-                        init_zero: bool = False,
-                        data_format: Text = 'channels_last',
-                        momentum: float = 0.99,
-                        epsilon: float = 1e-3,
-                        strategy: Text = None,
-                        name: Text = 'tpu_batch_normalization'):
-  """Performs a batch normalization followed by a non-linear activation.
+def build_batch_norm(is_training_bn: bool,
```
mingxingtan, line 24: I changed batch_normalization to build_batch_norm, because batch_normalization reads like a function that takes a tensor as input and run the batch norm, but here it actually just build the batch_norm_layer.

PR #451, opened 2020-05-26
fix clip boxes

fsx950223:


efficientdet/dataloader.py
```diff
@@ -237,6 +240,7 @@ def __call__(self, params):
     anchor_labeler = anchors.AnchorLabeler(input_anchors, params['num_classes'])
     example_decoder = tf_example_decoder.TfExampleDecoder()
```
mingxingtan, line 242: What does this tf.autograph.experimental.do_not_convert do?

fsx950223: I could debug the code.

mingxingtan: would it be slower?

fsx950223: > would it be slower?

No.

PR #466, opened 2020-06-02
add efficientdet keras model

fsx950223:


efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -25,64 +25,239 @@
...
 
-  def call(self, feats):
-    # @TODO: Implement this with keras logic
-    return legacy_arch.build_bifpn_layer(feats, self.feat_sizes, self)
+    return new_node
 
-  def get_config(self):
-    base_config = super(BiFPNLayer, self).get_config()
+  def call(self, feats):
+    # num_output_connections = [0 for _ in feats]
```
fsx950223, line 139: num_output_connections only for debug?

mingxingtan: yes, it was for debugging, and you don't need it here.

PR #486, opened 2020-06-07
add rw params

fsx950223:
Use ```model.params``` to keep track of params.
I will add it to saved model and ckpt path.
cc @mingxingtan


efficientdet/main.py
```diff
@@ -240,10 +241,16 @@ def _can_partition(spatial_dim):
       per_host_input_for_training=tf.estimator.tpu.InputPipelineConfig
       .PER_HOST_V2)
 
+  if FLAGS.model_dir is None:
```
mingxingtan, line 244: Maybe we can simple mark model_dir as required?

If users get used to specify model_dir=None or not specify it, then they will often have troubles to continue a stopped training job: when they restart the training job, model_dir will change and the job wouldn't start from the previously trained checkpoionts.

WDYT?

efficientdet/utils.py
```diff
@@ -31,6 +32,16 @@
 from tensorflow.python.tpu import tpu_function  # pylint:disable=g-direct-tensorflow-import
 # pylint: disable=logging-format-interpolation
 
+def rw_params(params):
+  os.makedirs(params['model_dir'], exist_ok=True)
+  params_path = os.path.join(params['model_dir'], 'model.params')
+  if os.path.exists(params_path):
+    content = open(params_path,'rb').read()
+    params = json.loads(content.decode(encoding="utf-8"))
```
mingxingtan, line 40: Should we use yaml for consistency?

efficientdet/main.py
fsx950223: I specific model dir by shell in my repo(Also use timestamp).
But for beginners, it's very convenient.
And I believe continue training is an advanced feature, they have ability to specific model dir when they need to restart training jobs.
Two strategies are goods for me.

efficientdet/utils.py
fsx950223: Good point.

Issue #488, opened 2020-06-07
Compilation failure on TPU


fishyboy:
During training efficientdet d7 on google colab tpu with my custom dataset, i encountered this problem:

I0607 16:13:53.142851 140561397208832 tpu_estimator.py:540] Starting outfeed thread controller.
INFO:tensorflow:Enqueue next (100) batch(es) of data to infeed.
I0607 16:13:57.123712 140563857274752 tpu_estimator.py:600] Enqueue next (100) batch(es) of data to infeed.
INFO:tensorflow:Dequeue next (100) batch(es) of data from outfeed.
I0607 16:13:57.124172 140563857274752 tpu_estimator.py:604] Dequeue next (100) batch(es) of data from outfeed.
ERROR:tensorflow:Error recorded from outfeed: From /job:worker/replica:0/task:0:
Bad hardware status: Queue is closed.
	 [[node OutfeedDequeueTuple (defined at /usr/local/lib/python3.6/dist-packages/tensorflow_core/python/framework/ops.py:1748) ]]

**ERROR:tensorflow:Error recorded from training_loop: From /job:worker/replica:0/task:0:
Compilation failure: Detected unsupported operations when trying to compile graph _functionalize_body_0[] on XLA_TPU_JIT: Const (No registered 'Const' OpKernel for XLA_TPU_JIT devices compatible with node node box_net/box-4-bn-7/batchnorm/add/y (defined at /usr/local/lib/python3.6/dist-packages/tensorflow_core/python/framework/ops.py:1748) 
	 (OpKernel was found, but attributes didn't match) Requested Attributes: _xla_inferred_shapes=[[]], dtype=DT_HALF, value=Tensor<type: half shape: [] values: 0.00100040436>, _device="/device:TPU_REPLICATED_CORE")node box_net/box-4-bn-7/batchnorm/add/y (defined at /usr/local/lib/python3.6/dist-packages/tensorflow_core/python/framework/ops.py:1748) 
	 [[LoopCond]]
	TPU compilation failed**
	 [[tpu_compile_succeeded_assert/_13283801156586978321/_2629]]
E0607 16:15:01.623807 140563857274752 error_handling.py:75] Error recorded from training_loop: From /job:worker/replica:0/task:0:
[...]

PR #490, opened 2020-06-08
enhance mixed precision

fsx950223:


efficientdet/hparams_config.py
```diff
@@ -222,8 +222,7 @@ def default_detection_configs():
   h.weight_decay = 4e-5
   # use horovod for multi-gpu training. If None, use TF default.
   h.strategy = None  # 'tpu', 'horovod', None
-  # precision: one of 'float32', 'mixed_float16', 'mixed_bfloat16'.
-  h.precision = None  # If None, use float32.
+  h.mixed_precision = False  # If False, use float32.
```
mingxingtan, line 225: Isn't precision is good enough? Do we need this extra flag? (it seems to introduce some extra code complexity)

efficientdet/aug/autoaugment.py
```diff
@@ -38,6 +37,10 @@
 except ImportError:
   from tensorflow.contrib import image as image_ops  # pylint: disable=g-import-not-at-top
 
+try:
+  import tensorflow_probability as tfp
+except ImportError:
+  import tensorflow.distributions as tfp
```
mingxingtan, line 43: Could you add a simple comment to these two comments? (which one is recommended by tf2?)

efficientdet/hparams_config.py
fsx950223: According to https://github.com/google/automl/issues/488.
TPUs don't support float16?

efficientdet/inference.py
```diff
@@ -162,10 +162,18 @@ def build_model(model_name: Text, inputs: tf.Tensor, **kwargs):
...
+    pp = 'float32'
   cls_outputs, box_outputs = utils.build_model_with_precision(
-      kwargs.get('precision', None), model_arch, inputs, False, model_name,
+      pp, model_arch, inputs, False, model_name,
       **kwargs)
-  if kwargs.get('precision', None):
+  if mixed_precision:
     # Post-processing has multiple places with hard-coded float32.
     # TODO(tanmingxing): Remove them once post-process can adpat to dtypes.
     cls_outputs = {k: tf.cast(v, tf.float32) for k, v in cls_outputs.items()}
```
fsx950223, line 179: Could remove them?

fsx950223:
I try to fix the tutorial. But my editor rewrite the 90% tutorial.
I'm afraid that I can't do it anymore.

PR #493, opened 2020-06-09
Allow min_score_thresh=0 during inference

drauh:
Fixes a bug I encountered when trying to set min_score_thresh=0 during inference; it was replaced by the default value (0.4).


mingxingtan:
Thanks for the PR!

PR #502, opened 2020-06-11
Add Yolo4 tricks to improve performance [WIP]

Ely-S:
Hi @mingxingtan

I'd like to implement some of the cheap tricks used in the  [yolo4  paper](https://arxiv.org/pdf/2004.10934.pdf) to increase EfficientDet's performance.

Here are some of the things I'm thinking of implementing and testing:

- [x] Class label smoothing
- [x] Mish activation
- [x] Hard swish activation from mobilenetv3. I want to see if it improves inference time.
- [x] DIOU NMS
- CIOU loss

Maybe you've thought of this before. Do you have any insights/suggestions concerning this line of work? Have any of these been tried for this project?

Thanks!

### DIFF
--- a/efficientdet/det_model_fn.py
+++ b/efficientdet/det_model_fn.py
@@ -165,7 +165,7 @@ def focal_loss(y_pred, y_true, alpha, gamma, label_smoothing, normalizer):
 
     # Get the cross_entropy for each entry
     ce = tf.keras.losses.categorical_crossentropy(
-        y_true, y_pred, from_logints=True, label_smoothing=label_smoothing)
+        y_true, y_pred, from_logits=True, label_smoothing=label_smoothing)
 
     pred_prob = tf.sigmoid(y_pred)
 
--- a/efficientdet/det_model_fn.py
+++ b/efficientdet/det_model_fn.py
@@ -157,7 +157,7 @@ def focal_loss(y_pred, y_true, alpha, gamma, label_smoothing, normalizer):
       and (1-alpha) to the loss from negative examples.
     gamma: A float32 scalar modulating loss from hard and easy examples.
     label_smoothing: Float in [0, 1]. If > `0` then smooth the labels.
-    normalzier: Divide loss by this value
+    normalizer: Divide loss by this value
   """
   with tf.name_scope('focal_loss'):
     alpha = tf.convert_to_tensor(alpha, dtype=y_pred.dtype)
--- a/efficientdet/det_model_fn.py
+++ b/efficientdet/det_model_fn.py
@@ -204,18 +204,6 @@ def focal_loss(logits, targets, alpha, gamma, normalizer):
   return weighted_loss
 
 
-def _classification_loss(cls_outputs,
-                         cls_targets,
-                         num_positives,
-                         alpha=0.25,
-                         gamma=2.0):
-  """Computes classification loss."""
-  normalizer = num_positives
-  classification_loss = focal_loss(cls_outputs, cls_targets, alpha, gamma,
-                                   normalizer)
-  return classification_loss
-
-
 def _box_loss(box_outputs, box_targets, num_positives, delta=0.1):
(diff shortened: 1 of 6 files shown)


efficientdet/utils.py
```diff
@@ -38,10 +38,14 @@ def activation_fn(features: tf.Tensor, act_type: Text):
     return tf.nn.swish(features)
   elif act_type == 'swish_native':
     return features * tf.sigmoid(features)
+  elif act_type == 'hard_swish':
```
mingxingtan, line 41: maybe just name it as hswish?

Ely-S: sure

efficientdet/det_model_fn.py
```diff
@@ -289,12 +286,14 @@ class and box losses from all levels.
...
                                         [bs, width, height, -1])
     box_targets_at_level = labels['box_targets_%d' % level]
-    cls_loss = _classification_loss(
-        cls_outputs[level],
-        cls_targets_at_level,
-        num_positives_sum,
-        alpha=params['alpha'],
-        gamma=params['gamma'])
+
+    cls_loss = focal_loss(
```
mingxingtan, line 290: focal_loss has some hard-coded numbers (like "positive_label_mask = tf.equal(targets, 1.0)"), you may want to update that?

You can also borrow some ideas or reuse the tfa.focal_loss? 

@fsx950223 seems to be familiar to tfa (tf addons)

Ely-S: Sure, I can update the focal loss function to be more like tfa focal_loss.

I would definitely appreciate @fsx950223 's insights.

Ely-S: I updated the focal loss function by adapting logic from the tfa version. Do you like this new version better? If so, can you please check the algebra? If I made a mistake in the math it would be pretty hard to catch at runtime. Thanks

Ely-S: For reference: https://github.com/tensorflow/addons/blob/v0.10.0/tensorflow_addons/losses/focal_loss.py#L90

fsx950223:
I suggest use tensorflow addons modules(such as mish, focal loss) instead of implement it by yourself and add some test cases for keep unify.

Ely-S:
@fsx950223 I would do that, but it would break compatibility with TF 1.x

mingxingtan:
@Ely-S @fsx950223 I am going to merge this PR, followed with a PR (added tests, fixed a few format issues, and merged some internal changes).

Ely-S:
Thanks for merging @mingxingtan

Zzh-tju:
@mingxingtan I've been interested in EfficientDet for a while. I noticed that you might want to use DIoU-NMS. It is true that if we directly replace original NMS with DIoU-NMS, it will slow down the inference speed. Our latest works Cluster-NMS and its combination with various NMS variants can accelerate the NMS process and improve both AP and AR. Codes and description of usage are in (https://github.com/Zzh-tju/CIoU), (https://github.com/Zzh-tju/DIoU-SSD-pytorch), and paper (https://arxiv.org/abs/2005.03572)

Wish these will help you.

mingxingtan:
Hi @Zzh-tju , thanks for the information. Do you want to prepare a PR to add your cluster-nms? (It is in numpy, so should be easy to implement like these code [here](https://github.com/google/automl/blob/7ab026b5a40a5583fbce39deb268298f0e6bfafc/efficientdet/anchors.py#L217))?  I would be happy to merge your PR and give it a try.

Zzh-tju:
@mingxingtan Your NMS function is CPU mode？If you can transform to tf, it could be accelerated by GPU as long as tf can do the same matrix operation as Pytorch does.

Ely-S:
@Zzh-tju  There is an implementation in tensorflow and one in numpy depending on the `disable_pyfun` internal parameter.

Issue #378, opened 2020-05-07
Add augmentation during the training


bessszilard:
I successfully trained the model, but I wanted to increase its robustness by adding augmentation. I tried to add augmentation with this command:

```bash
!python main.py --mode=train_and_eval \
    --training_file_pattern={train_file} \
    --validation_file_pattern={eval_file} \
    --model_name={MODEL} \
    --model_dir=/content/tmp/model_dir/{MODEL}-finetune \
    --ckpt={MODEL} \
    --train_batch_size=1 --eval_batch_size=1 \
    --eval_samples={images_per_eval}  \
    --num_examples_per_epoch={images_per_epoch} --num_epochs=20 \
    --hparams="num_classes=4,moving_average_decay=0,autoaugment_policy=v0" \
    --use_tpu=False
```

But I received this error:
```
INFO:tensorflow:Running local_init_op.
I0507 10:36:35.495698 140318379759488 session_manager.py:505] Running local_init_op.
INFO:tensorflow:Done running local_init_op.
I0507 10:36:35.882802 140318379759488 session_manager.py:508] Done running local_init_op.
INFO:tensorflow:Calling checkpoint listeners before saving checkpoint 0...
I0507 10:36:48.344113 140318379759488 basic_session_run_hooks.py:614] Calling checkpoint listeners before saving checkpoint 0...
INFO:tensorflow:Saving checkpoints for 0 into /content/tmp/model_dir/efficientdet-d0-finetune/model.ckpt.
I0507 10:36:48.358273 140318379759488 basic_session_run_hooks.py:618] Saving checkpoints for 0 into /content/tmp/model_dir/efficientdet-d0-finetune/model.ckpt.
INFO:tensorflow:Calling checkpoint listeners after saving checkpoint 0...
I0507 10:36:51.328609 140318379759488 basic_session_run_hooks.py:626] Calling checkpoint listeners after saving checkpoint 0...
Fatal Python error: Segmentation fault

Thread 0x00007f9d97bf8700 (most recent call first):
  File "/usr/lib/python3.6/threading.py", line 295 in wait
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/summary/writer/event_file_writer.py", line 266 in get
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/summary/writer/event_file_writer.py", line 209 in run
  File "/usr/lib/python3.6/threading.py", line 916 in _bootstrap_inner
[...]

Issue #161, opened 2020-04-06
autoaugment_policy crash


chentairan:
Hi,

When I try to use data augmentation, set `h.autoaugment_policy = 'v1'` or `h.autoaugment_policy = 'v2'`, running `main.py` crash. 

Here is the error information:

```
Traceback (most recent call last):
  File "main.py", line 390, in <module>
    tf.app.run(main)
  File "/home/mist/miniconda3/lib/python3.7/site-packages/tensorflow_core/python/platform/app.py", line 40, in run
    _run(main=main, argv=argv, flags_parser=_parse_flags_tolerate_undef)
  File "/home/mist/miniconda3/lib/python3.7/site-packages/absl/app.py", line 299, in run
    _run_main(main, args)
  File "/home/mist/miniconda3/lib/python3.7/site-packages/absl/app.py", line 250, in _run_main
    sys.exit(main(argv))
  File "main.py", line 255, in main
    FLAGS.train_batch_size))
  File "/home/mist/miniconda3/lib/python3.7/site-packages/tensorflow_estimator/python/estimator/tpu/tpu_estimator.py", line 3035, in train
    rendezvous.raise_errors()
  File "/home/mist/miniconda3/lib/python3.7/site-packages/tensorflow_estimator/python/estimator/tpu/error_handling.py", line 143, in raise_errors
    six.reraise(typ, value, traceback)
  File "/home/mist/miniconda3/lib/python3.7/site-packages/six.py", line 703, in reraise
    raise value
  File "/home/mist/miniconda3/lib/python3.7/site-packages/tensorflow_estimator/python/estimator/tpu/tpu_estimator.py", line 3030, in train
    saving_listeners=saving_listeners)
  File "/home/mist/miniconda3/lib/python3.7/site-packages/tensorflow_estimator/python/estimator/estimator.py", line 374, in train
    loss = self._train_model(input_fn, hooks, saving_listeners)
  File "/home/mist/miniconda3/lib/python3.7/site-packages/tensorflow_estimator/python/estimator/estimator.py", line 1164, in _train_model
    return self._train_model_default(input_fn, hooks, saving_listeners)
  File "/home/mist/miniconda3/lib/python3.7/site-packages/tensorflow_estimator/python/estimator/estimator.py", line 1198, in _train_model_default
    saving_listeners)
  File "/home/mist/miniconda3/lib/python3.7/site-packages/tensorflow_estimator/python/estimator/estimator.py", line 1497, in _train_with_estimator_spec
[...]

PR #509, opened 2020-06-12
Force to use cpu for depthwise_conv2d

bessszilard:
* to fix https://github.com/google/automl/issues/378
* based on https://github.com/google/automl/issues/161#issuecomment-612966054


bessszilard:
@googlebot I signed it!

PR #515, opened 2020-06-15
Fixed test_model_output of keras implementation

d61h6k4:
Changed the format of intermidiate representaiton of features from dict
to list (Invoking keras layer only accepts Tensor/tuple/list)

Converted the code which adds resample feature maps to backbone features
to the keras layer ResampleFeatureAdder (needed to compile keras model)

Converted function fuse_features to keras layer FuseFeature

Added methods which test that the layers of the keras model are compilable.

Changed representation of model in test_model_output function from
keras model to chain of keras layers because the output of the keras model is
different to the output of the previous model. (can this be the cause of
keras model not using session graph?, so its not using the set random seed?)

### DIFF
--- a/efficientdet/keras/efficientdet_arch_keras.py
+++ b/efficientdet/keras/efficientdet_arch_keras.py
@@ -27,6 +27,71 @@ from backbone import efficientnet_builder
 from keras import utils_keras
 
 
+class FuseFeatures(tf.keras.layers.Layer):
+  """Fuse features from different resolutions and return a weighted sum."""
+  def __init__(self, inputs_offsets, weight_method, fpn_num_filters, **kwargs):
+    super(FuseFeatures, self).__init__(**kwargs)
+
+    self.inputs_num = len(inputs_offsets)
+    self.inputs_offsets = inputs_offsets
+    self.weight_method = weight_method
+    self.fpn_num_filters = fpn_num_filters
+    self.edge_weights = []
+
+    if self.weight_method == 'attn':
+      self._add_wsm('ones')
+    elif self.weight_method == 'fastattn':
+      self._add_wsm('ones')
+    elif self.weight_method == 'channel_attn':
+      num_filters = int(self.fpn_num_filters)
+      self._add_wsm(lambda shape, dtype: tf.ones(shape, dtype=dtype))
+    elif self.weight_method == 'channel_fastattn':
+      num_filters = int(self.fpn_num_filters)
+      self._add_wsm(lambda shape, dtype: tf.ones(shape, dtype=dtype))
+    elif self.weight_method == 'sum':
+      # no weight for sum
+      pass
+    else:
+      raise ValueError('unknown weight_method {}'.format(weight_method))
+
+  def _add_wsm(self, initializer):
+    for i in range(self.inputs_num):
+      if i == 0:
+        name = 'WSM'
+      else:
+        name = 'WSM_{}'.format(i)
+      self.edge_weights.append(
+          self.add_weight(initializer=initializer, name=name, trainable=True, dtype=self.dtype))
+
+  def call(self, nodes):
+    if self.weight_method == 'attn':
+      normalized_weights = tf.nn.softmax(tf.stack(self.edge_weights))
+      nodes = tf.stack(nodes, axis=-1)
+      new_node = tf.reduce_sum(nodes * normalized_weights, -1)
+    elif self.weight_method == 'fastattn':
+      weights_sum = tf.math.add_n(self.edge_weights)
+      nodes = [
+          nodes[i] * self.edge_weights[i] / (weights_sum + 0.0001)
+          for i in range(self.inputs_num)
+      ]
+      new_node = tf.math.add_n(nodes)
+    elif self.weight_method == 'channel_attn':
+      normalized_weights = tf.nn.softmax(tf.stack(self.edge_weights, -1), axis=-1)
+      nodes = tf.stack(nodes, axis=-1)
+      new_node = tf.reduce_sum(nodes * normalized_weights, -1)
+    elif self.weight_method == 'channel_fastattn':
+      weights_sum = tf.math.add_n(self.edge_weights)
+      nodes = [
+          nodes[i] * self.edge_weights[i] / (weights_sum + 0.0001)
+          for i in range(self.inputs_num)
+      ]
+      new_node = tf.math.add_n(nodes)
+    elif self.weight_method == 'sum':
+      new_node = tf.math.add_n(nodes)
+
+    return new_node
+
+
 class FNode(tf.keras.layers.Layer):
   """A Keras Layer implementing BiFPN Node."""
 
@@ -63,77 +128,10 @@ class FNode(tf.keras.layers.Layer):
     self.data_format = data_format
     self.weight_method = weight_method
     self.conv_bn_act_pattern = conv_bn_act_pattern
-    self.resample_feature_maps = []
-    self.op_after_combines = []
-    self.vars = []
 
-  def fuse_features(self, nodes):
-    """Fuse features from different resolutions and return a weighted sum.
+    self.fuse_features = FuseFeatures(inputs_offsets, weight_method, fpn_num_filters, dtype=self.dtype)
 
-    Args:
-      nodes: a list of tensorflow features at different levels
-
-    Returns:
-      A tensor denoting the fused feature.
-    """
-    dtype = nodes[0].dtype
-
-    if self.weight_method == 'attn':
-      edge_weights = []
-      for var in self.vars:
-        var = tf.cast(var, dtype=dtype)
-        edge_weights.append(var)
-      normalized_weights = tf.nn.softmax(tf.stack(edge_weights))
-      nodes = tf.stack(nodes, axis=-1)
-      new_node = tf.reduce_sum(nodes * normalized_weights, -1)
-    elif self.weight_method == 'fastattn':
-      edge_weights = []
-      for var in self.vars:
-        var = tf.cast(var, dtype=dtype)
-        edge_weights.append(var)
-      weights_sum = tf.add_n(edge_weights)
-      nodes = [
-          nodes[i] * edge_weights[i] / (weights_sum + 0.0001)
-          for i in range(len(nodes))
-      ]
-      new_node = tf.add_n(nodes)
-    elif self.weight_method == 'channel_attn':
-      edge_weights = []
-      for var in self.vars:
-        var = tf.cast(var, dtype=dtype)
-        edge_weights.append(var)
-      normalized_weights = tf.nn.softmax(tf.stack(edge_weights, -1), axis=-1)
-      nodes = tf.stack(nodes, axis=-1)
-      new_node = tf.reduce_sum(nodes * normalized_weights, -1)
-    elif self.weight_method == 'channel_fastattn':
-      edge_weights = []
-      for var in self.vars:
-        var = tf.cast(var, dtype=dtype)
-        edge_weights.append(var)
-
-      weights_sum = tf.add_n(edge_weights)
-      nodes = [
-          nodes[i] * edge_weights[i] / (weights_sum + 0.0001)
-          for i in range(len(nodes))
-      ]
-      new_node = tf.add_n(nodes)
-    elif self.weight_method == 'sum':
-      new_node = tf.add_n(nodes)
-    else:
-      raise ValueError('unknown weight_method {}'.format(self.weight_method))
-
-    return new_node
-
-  def _add_wsm(self, initializer):
-    for i, _ in enumerate(self.inputs_offsets):
-      if i == 0:
-        name = 'WSM'
-      else:
-        name = 'WSM_{}'.format(i)
-      self.vars.append(
-          self.add_weight(initializer=initializer, name=name, trainable=True))
-
-  def build(self, feats_shape):
+    self.resample_feature_maps = {}
     for idx, input_offset in enumerate(self.inputs_offsets):
       resample_feature_map = ResampleFeatureMap(self.new_node_height,
                                                 self.new_node_width,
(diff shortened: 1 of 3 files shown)


mingxingtan:
This is great. Thank you, @d61h6k4 
I will review in late today.

efficientdet/keras/efficientdet_arch_keras_test.py
```diff
@@ -170,7 +382,7 @@ def build_model(self, keras=False):
         efficientdet_arch_keras.build_class_and_box_outputs(inputs, config)
       return [n.name for n in tf.global_variables()]
 
-  def test_graph_variables_name_compatibility(self):
+  def _test_graph_variables_name_compatibility(self):
```
fsx950223, line 385: Why do you change the test case

d61h6k4: In this PR I disabled all name compatibility tests and focused on getting same outputs with "previous" model and checked that I can compile the keras model.

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -27,6 +27,71 @@
 from keras import utils_keras
 
 
+class FuseFeatures(tf.keras.layers.Layer):
```
fsx950223, line 30: What's meaning of change it to keras layers?

efficientdet/keras/efficientdet_arch_keras_test.py
fsx950223: > In this PR I disabled all name compatibility tests and focused on getting same outputs with "previous" model and checked that I can compile the keras model.

The mode could get same output now.

efficientdet/keras/utils_keras.py
```diff
@@ -18,7 +18,14 @@
 
 from typing import Text
 import tensorflow as tf
-import utils
+
+
+def batch_norm_class(is_training, strategy=None):
+  if is_training and strategy == 'horovod':
+    return tf.keras.layers.experimental.SyncBatchNormalization
```
fsx950223, line 25: I don't understand.

efficientdet/keras/efficientdet_arch_keras_test.py
d61h6k4: That's great

efficientdet/keras/utils_keras.py
d61h6k4: I got warning on tf.add_to_collection from tensorflow, so to fix it I added this function

fsx950223:
I suggest that you could implement a subclass keras model which is friendly for debugging.Don't need to change layers.
I believe the current keras graph model is good enough.

efficientdet/keras/efficientdet_arch_keras.py
d61h6k4: When I loaded the model as keras.Model, I got the error `tenorflow.python.framework.errors_impl.FailedPreconditionError`
I was able to localize it before calling the `fuse_feature` function, so to fix this, I created a layer and started covering it with tests (this was the original reason).

d61h6k4:
I fixed variables name compatibility tests, converted back FuseFeatures to the method and removed duplicate functions from utils_keras.

d61h6k4:
When I tried to restore the Keras model from the checkpoint (I downloaded efficiendet-d0 ckpt) I got 
` ValueError: Shapes (1, 1, 8, 32) and (1, 1, 32, 8) are incompatible`.
Do you know what the tensor can it be?

d61h6k4:
With the last changes, I am able to restore the `keras.Model` from the checkpoint.

```
import keras.efficientdet_arch_keras as efficientdet_arch
import tensorflow as to

model = efficientdet_arch.efficientdet("efficientdet-d0")
optimizer = tf.keras.optimizers.Adam()
checkpoint = tf.train.Checkpoint(optimizer=optimizer, model=model)
status = checkpoint.restore(tf.train.latest_checkpoint("models/efficientdet-d0"))
status.assert_consumed()
```

mingxingtan:
Wow, this is great, @d61h6k4 

Are you able to do some inference using keras models as well?

efficientdet/backbone/efficientnet_model.py
```diff
@@ -248,10 +260,27 @@ def _build(self):
         padding='same',
         data_format=self._data_format,
         use_bias=False)
+
+
+    if self._block_args.fused_conv:
+      bn0_name = "tpu_batch_normalization"
```
mingxingtan, line 266: Do you need these changes? Since they are already keras layers and it works well with existing code?

efficientdet/keras/utils_keras.py
```diff
@@ -52,11 +52,11 @@ def build_batch_norm(is_training_bn: bool,
   axis = 1 if data_format == 'channels_first' else -1
   batch_norm_class = utils.batch_norm_class(is_training_bn, strategy)
   bn_layer = batch_norm_class(axis=axis,
-                              momentum=momentum,
-                              epsilon=epsilon,
-                              center=True,
-                              scale=True,
-                              gamma_initializer=gamma_initializer,
-                              name=name)
+                            momentum=momentum,
```
mingxingtan, line 55: Revert these changes.

You may want to use the linter to check format: https://github.com/google/automl/blob/master/efficientdet/g3doc/faq.md#21-how-can-i-format-my-code-for-prs

efficientdet/backbone/efficientnet_model.py
d61h6k4: Yes, I need them, because without them variable names are different. 
When keras model created in the context of `tf.Graph().as_default` you can go without this naming and variables names will be equal with tensorflow 1 model version, but when keras model created without this context, variables names are different. I changed the `test_model_output_variables` test to handle it.

efficientdet/keras/utils_keras.py
d61h6k4: Thanks, I will apply the linter to check format.

d61h6k4:
> Wow, this is great, @d61h6k4
> 
> Are you able to do some inference using keras models as well?

Yes, I able to inference and I checked that predictions on `tf.ones([1, 512, 512, 3])` before restoring from checkpoint and after restoring are different, but nothing more difficult.

efficientdet/backbone/efficientnet_model.py
fsx950223: You shouldn't remove graph test case.

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -406,7 +389,7 @@ def __init__(self,
 
       bn_per_level = {}
       for level in range(self.min_level, self.max_level + 1):
-        bn_per_level[level] = utils_keras.build_batch_norm(
+        bn_per_level[str(level)] = utils_keras.build_batch_norm(
```
fsx950223, line 392: Why I need this?

fsx950223:
```
from keras.efficientdet_arch_keras import efficientdet
import tensorflow as tf

model = efficientdet('efficientdet-d0')
ckpt = tf.train.load_checkpoint('checkpoint_path')
for var in model.variables:
  var.assign(ckpt.get_tensor(var.name.split(':')[0]))
model.save_weights('./test.h5', save_format='h5')
model.load_weights('./test.h5')
```
Existing code could read and write weights by the way.

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -696,6 +681,62 @@ def call(self, feats):
     return feats
 
 
+class ResampleFeatureAdder(tf.keras.layers.Layer):
```
fsx950223, line 684: I won't approve such modifies that just wrapper it with a keras layer and change code positions.

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -161,19 +88,75 @@ def build(self, feats_shape):
                                            self.data_format,
                                            self.strategy,
                                            name='op_after_combine{}'.format(
-                                               len(feats_shape)))
-    self.built = True
+                                               len(input_shape)))
+
+    if self.weight_method == 'attn':
```
fsx950223, line 93: I won't approve such modifies that just wrapper it with a keras layer and change code positions.

efficientdet/keras/efficientdet_arch_keras_test.py
```diff
@@ -25,6 +27,199 @@
...
+                              feat_sizes=feat_sizes,
+                              min_level=config.min_level,
+                              max_level=config.max_level,
+                              data_format=config.data_format)
+    except ValueError as err:
+      self.assertFalse(True, msg=repr(err))
+    self.assertEqual(len(preds), 5, "P3-P7")
+
+  def test_fnode_compile(self):
+    config = hparams_config.get_efficientdet_config("efficientdet-d0")
```
fsx950223, line 67: Don't need compile tests.

efficientdet/keras/efficientdet_arch_keras.py
d61h6k4: This change allows me to save the model, I found this trick here: https://github.com/tensorflow/tensorflow/issues/36916#issuecomment-640395999

P.S. I mentioned it in commit message.

d61h6k4:
> ```
> from keras.efficientdet_arch_keras import efficientdet
> import tensorflow as tf
> 
> model = efficientdet('efficientdet-d0')
> ckpt = tf.train.load_checkpoint('checkpoint_path')
> for var in model.variables:
>   var.assign(ckpt.get_tensor(var.name.split(':')[0]))
> model.save_weights('./test.h5', save_format='h5')
> model.load_weights('./test.h5')
> ```
> 
> Existing code could read and write weights by above process.

If that means that we can restore the keras model from the pre-trained checkpoint, that I think we don't need this PR

fsx950223:
> > ```
> > from keras.efficientdet_arch_keras import efficientdet
> > import tensorflow as tf
> > 
> > model = efficientdet('efficientdet-d0')
> > ckpt = tf.train.load_checkpoint('checkpoint_path')
> > for var in model.variables:
> >   var.assign(ckpt.get_tensor(var.name.split(':')[0]))
> > model.save_weights('./test.h5', save_format='h5')
> > model.load_weights('./test.h5')
> > ```
> > 
> > 
> > Existing code could read and write weights by above process.
> 
> If that means that we can restore the keras model from the pre-trained checkpoint, that I think we don't need this PR



> > ```
> > from keras.efficientdet_arch_keras import efficientdet
> > import tensorflow as tf
> > 
> > model = efficientdet('efficientdet-d0')
> > ckpt = tf.train.load_checkpoint('checkpoint_path')
> > for var in model.variables:
> >   var.assign(ckpt.get_tensor(var.name.split(':')[0]))
> > model.save_weights('./test.h5', save_format='h5')
> > model.load_weights('./test.h5')
> > ```
> > 
> > 
> > Existing code could read and write weights by above process.
> 
> If that means that we can restore the keras model from the pre-trained checkpoint, that I think we don't need this PR

Yes, you could submit a standalone PR to fix tf format checkpoint.

efficientdet/keras/efficientdet_arch_keras.py
d61h6k4: I created this layer cause I don't know how I can implement it without it. If you know how to create it without keras layer, please share your ideas.

fsx950223: Exist code has implemented it.

efficientdet/keras/efficientdet_arch_keras.py
fsx950223: Change dict to list should fix the bug?

efficientdet/keras/efficientdet_arch_keras.py
d61h6k4: The reason of this PR is the existing code didn't work for me, especially I couldn't inference with existing keras model.
> You should add resample_feature_adder variable scope to tf1.x for unify variables' names if you want to add the layer.
you shouldn't, I fixed it, that why I am able to restore the model from the checkpoint.

efficientdet/keras/efficientdet_arch_keras.py
d61h6k4: yes, but there he explains why to touch `__dict__` is problematic.

efficientdet/keras/efficientdet_arch_keras.py
fsx950223: It may just your problem.
I can't meet any bug when I run test cases.

efficientdet/keras/efficientdet_arch_keras.py
fsx950223: > yes, but there he explains why to touch `__dict__` is problematic.

Good Observation.

efficientdet/keras/efficientdet_arch_keras.py
d61h6k4: > It may just your problem.

agree

> I can't meet any bug when I run test cases.

sure, but in the current implementation, keras model ran only under the `tf.Session` and with `sess.run(tf.global_variables_initializer())`, but in my case, I would like to run them without session.

fsx950223: > > It may just your problem.
> 
> agree
> 
> > I can't meet any bug when I run test cases.
> 
> sure, but in the current implementation, keras model ran only under the `tf.Session` and with `sess.run(tf.global_variables_initializer())`, but in my case, I would like to run them without session.

```
from keras.efficientdet_arch_keras import efficientdet
import tensorflow as tf

model = efficientdet('efficientdet-d0')
ckpt = tf.train.load_checkpoint('checkpoint_path')
for var in model.variables:
  var.assign(ckpt.get_tensor(var.name.split(':')[0]))
model.save_weights('./test.h5', save_format='h5')
model.load_weights('./test.h5')
```
Sure, it doesn't need Session anymore.

PR #520, opened 2020-06-16
fix keras tests

fsx950223:
Finish keras model refact.

### DIFF
--- a/efficientdet/keras/efficientdet_arch_keras_test.py
+++ b/efficientdet/keras/efficientdet_arch_keras_test.py
@@ -20,32 +20,49 @@ import tensorflow.compat.v1 as tf
 import efficientdet_arch as legacy_arch
 import hparams_config
 from keras import efficientdet_arch_keras
-import functools
-import utils
 
 SEED = 111111
 
 
 class KerasTest(tf.test.TestCase):
 
-  def _test_model_output(self):
-    # TODO(fsx950223): Fix the test case
+  def test_backbone(self):
     inputs_shape = [1, 512, 512, 3]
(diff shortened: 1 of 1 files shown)


efficientdet/keras/efficientdet_arch_keras_test.py
```diff
@@ -28,12 +28,30 @@
...
+      feats = tf.ones(inputs_shape)
+      config = hparams_config.get_efficientdet_config('efficientdet-d0')
+      tf.random.set_random_seed(SEED)
+      feats = legacy_arch.build_backbone(feats, config)
+      sess.run(tf.global_variables_initializer())
+      feats2 = sess.run(feats)
+    for key in list(feats.keys()):
+      self.assertAllEqual(feats1[key], feats2[key])
+
   def _test_model_output(self):
```
fsx950223, line 50: Submodules could get same results with tf1.x model.
The difference may be caused by Keras model.
The test case could be removed.
Please verify it.

efficientdet/keras/efficientdet_arch_keras_test.py
```diff
@@ -177,9 +195,9 @@ def test_graph_variables_name_compatibility(self):
     self.assertEqual(legacy_names, keras_names)
 
   def test_output(self):
-    config = hparams_config.get_efficientdet_config()
+    config = hparams_config.get_efficientdet_config('efficientdet-d0')
     inputs_shape = [1, 512, 512, 3]
-
+    config.max_level = config.min_level + 1
```
fsx950223, line 200: Reduce memory usage.

PR #528, opened 2020-06-18
fix multi training value bug

fsx950223:


efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -217,7 +217,7 @@ def call(self, new_node):
     if not self.conv_bn_act_pattern:
       new_node = utils.activation_fn(new_node, self.act_type)
     new_node = self.conv_op(new_node)
-    new_node = self.bn(new_node, self.is_training_bn)
+    new_node = self.bn(new_node, training=self.is_training)
```
fsx950223, line 220: Cause multi training values bug.

efficientdet/keras/efficientdet_arch_keras_test.py
```diff
@@ -100,23 +100,32 @@ def test_build_feature_network(self):
       self.assertAllEqual(new_feats1[i], new_feats2[i])
 
   def test_model_variables(self):
+    feats = tf.ones([1, 512, 512, 3])
```
fsx950223, line 103: Verify keras model variables in eager mode.

PR #522, opened 2020-06-17
A followup to #502 and merge some internal changes.

mingxingtan:
Important changes:
1. Changed config to allow hierarchical dict or configs.

Minor changes:
2. Deleted some unless configs.
3. Added tests for focal loss.
4. Fixed some format and coding style issues.

PR #532, opened 2020-06-20
test label smoothing

Ely-S:
Hi @mingxingtan , I've been travelling this week and didn't get the chance to follow up on the comments on PRs #502 and #522 earlier. 

Here is a test for label smoothing. Please lmk if there is anything else to change or add. Thanks

### DIFF
--- a/efficientdet/det_model_fn.py
+++ b/efficientdet/det_model_fn.py
@@ -142,8 +142,8 @@ def learning_rate_schedule(params, global_step):
   raise ValueError('unknown lr_decay_method: {}'.format(lr_decay_method))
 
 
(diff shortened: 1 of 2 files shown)


mingxingtan:
Looks great. Thanks!
I tried label smoothing 0.1, but looks like it doesn't help accuracy. Is that because we are using binary classification?

Ely-S:
I did a bit of research and couldn't find a good explanation for that. Yolo3 also uses binary classification to predict one-hot labels and label smoothing benefited it by 0.64 mAP (https://arxiv.org/pdf/1902.04103.pdf)

PR #536, opened 2020-06-22
add train script

fsx950223:
Could you test tf syncbatchnorm, I will use it in keras model? @mingxingtan
I'm not familiar with TPU, could you add tf2 TPU logic?

TF version: 2.3+

### DIFF
--- a/efficientdet/dataloader.py
+++ b/efficientdet/dataloader.py
@@ -356,7 +356,7 @@ class InputReader(object):
 
     # Prefetch data from files.
     def _prefetch_dataset(filename):
-      dataset = tf.data.TFRecordDataset(filename, 'GZIP').prefetch(1)
+      dataset = tf.data.TFRecordDataset(filename).prefetch(1)
       return dataset
 
     dataset = dataset.apply(
--- a/efficientdet/dataloader.py
+++ b/efficientdet/dataloader.py
@@ -14,7 +14,7 @@
 # ==============================================================================
 """Data loader and processing."""
 from absl import logging
-import tensorflow.compat.v1 as tf
+import tensorflow as tf
 
 import anchors
 import utils
@@ -83,7 +83,7 @@ class InputProcessor(object):
                  self._output_size)
 
     # Select a random scale factor.
-    random_scale_factor = tf.random_uniform([], scale_min, scale_max)
+    random_scale_factor = tf.random.uniform([], scale_min, scale_max)
     scaled_y = tf.cast(random_scale_factor * target_size[0], tf.int32)
     scaled_x = tf.cast(random_scale_factor * target_size[1], tf.int32)
 
@@ -100,8 +100,8 @@ class InputProcessor(object):
     scaled_width = tf.cast(width * image_scale, tf.int32)
     offset_y = tf.cast(scaled_height - self._output_size[0], tf.float32)
     offset_x = tf.cast(scaled_width - self._output_size[1], tf.float32)
-    offset_y = tf.maximum(0.0, offset_y) * tf.random_uniform([], 0, 1)
-    offset_x = tf.maximum(0.0, offset_x) * tf.random_uniform([], 0, 1)
+    offset_y = tf.maximum(0.0, offset_y) * tf.random.uniform([], 0, 1)
+    offset_x = tf.maximum(0.0, offset_x) * tf.random.uniform([], 0, 1)
     offset_y = tf.cast(offset_y, tf.int32)
     offset_x = tf.cast(offset_x, tf.int32)
     self._image_scale = image_scale
@@ -126,7 +126,7 @@ class InputProcessor(object):
 
   def resize_and_crop_image(self, method=tf.image.ResizeMethod.BILINEAR):
     """Resize input image and crop it to the self._output dimension."""
-    scaled_image = tf.image.resize_images(
+    scaled_image = tf.image.resize(
         self._image, [self._scaled_height, self._scaled_width], method=method)
     scaled_image = scaled_image[self._crop_offset_y:self._crop_offset_y +
                                 self._output_size[0],
@@ -333,7 +333,7 @@ class InputReader(object):
 
         source_id = tf.where(
             tf.equal(source_id, tf.constant('')), '-1', source_id)
-        source_id = tf.string_to_number(source_id)
+        source_id = tf.strings.to_number(source_id)
 
         # Pad groundtruth data for evaluation.
         image_scale = input_processor.image_scale_to_original
@@ -356,7 +356,7 @@ class InputReader(object):
 
     # Prefetch data from files.
     def _prefetch_dataset(filename):
-      dataset = tf.data.TFRecordDataset(filename).prefetch(1)
+      dataset = tf.data.TFRecordDataset(filename, 'GZIP').prefetch(1)
       return dataset
 
     dataset = dataset.apply(
@@ -382,7 +382,15 @@ class InputReader(object):
               batch_size,
           ]), [batch_size, 1])
 
+      if params['data_format'] == 'channels_first':
+        images = tf.transpose(images, [0, 3, 1, 2])
+
       for level in range(params['min_level'], params['max_level'] + 1):
+        if params['data_format'] == 'channels_first':
+          labels['cls_targets_%d' % level] = tf.transpose(
+            labels['cls_targets_%d' % level], [0, 3, 1, 2])
+          labels['box_targets_%d' % level] = tf.transpose(
+            labels['box_targets_%d' % level], [0, 3, 1, 2])
         labels['cls_targets_%d' % level] = cls_targets[level]
         labels['box_targets_%d' % level] = box_targets[level]
       # Concatenate groundtruth annotations to a tensor.
--- a/efficientdet/det_model_fn.py
+++ b/efficientdet/det_model_fn.py
@@ -237,11 +237,6 @@ def detection_loss(cls_outputs, box_outputs, labels, params):
   box_losses = []
   box_iou_losses = []
   for level in levels:
-    if params['data_format'] == 'channels_first':
-      labels['cls_targets_%d' % level] = tf.transpose(
-          labels['cls_targets_%d' % level], [0, 3, 1, 2])
-      labels['box_targets_%d' % level] = tf.transpose(
-          labels['box_targets_%d' % level], [0, 3, 1, 2])
     # Onehot encoding for classification labels.
     cls_targets_at_level = tf.one_hot(labels['cls_targets_%d' % level],
                                       params['num_classes'])
@@ -436,8 +431,6 @@ def _model_fn(features, labels, mode, params, model, variable_filter_fn=None):
   """
   utils.image('input_image', features)
   training_hooks = []
-  if params['data_format'] == 'channels_first':
-    features = tf.transpose(features, [0, 3, 1, 2])
 
   def _model_outputs(inputs):
     # Convert params (dict) to Config for easier access.
--- a/efficientdet/keras/efficientdet_keras_test.py
+++ b/efficientdet/keras/efficientdet_keras_test.py
@@ -66,20 +66,28 @@ class EfficientDetKerasTest(tf.test.TestCase):
       sess.run(tf.global_variables_initializer())
       legacy_class_out, legacy_box_out = sess.run(feats)
     for i in range(3, 8):
-      self.assertAllClose(
-          keras_class_out[i - 3], legacy_class_out[i], rtol=1e-4, atol=1e-4)
-      self.assertAllClose(
-          keras_box_out[i - 3], legacy_box_out[i], rtol=1e-4, atol=1e-4)
+      self.assertAllClose(keras_class_out[i - 3],
+                          legacy_class_out[i],
+                          rtol=1e-4,
+                          atol=1e-4)
+      self.assertAllClose(keras_box_out[i - 3],
+                          legacy_box_out[i],
+                          rtol=1e-4,
+                          atol=1e-4)
 
     feats = tf.ones(inputs_shape)
     model = efficientdet_keras.EfficientDetNet(config=config)
     model.load_weights(tmp_ckpt)
     eager_class_out, eager_box_out = model(feats)
     for i in range(3, 8):
-      self.assertAllClose(
-        eager_class_out[i - 3], legacy_class_out[i], rtol=1e-4, atol=1e-4)
-      self.assertAllClose(
-        eager_box_out[i - 3], legacy_box_out[i], rtol=1e-4, atol=1e-4)
+      self.assertAllClose(eager_class_out[i - 3],
+                          legacy_class_out[i],
+                          rtol=1e-4,
+                          atol=1e-4)
+      self.assertAllClose(eager_box_out[i - 3],
+                          legacy_box_out[i],
+                          rtol=1e-4,
+                          atol=1e-4)
 
   def test_build_feature_network(self):
     config = hparams_config.get_efficientdet_config('efficientdet-d0')
@@ -199,8 +207,7 @@ class EfficientDetVariablesNamesTest(tf.test.TestCase):
         legacy_inputs[i] = keras_inputs[-1]
 
       if keras:
-        efficientdet_keras.build_class_and_box_outputs(
-            keras_inputs, config)
+        efficientdet_keras.build_class_and_box_outputs(keras_inputs, config)
       else:
         legacy_arch.build_class_and_box_outputs(legacy_inputs, config)
       return [n.name for n in tf.global_variables()]
--- a/efficientdet/keras/efficientdet_keras_test.py
+++ b/efficientdet/keras/efficientdet_keras_test.py
@@ -66,28 +66,20 @@ class EfficientDetKerasTest(tf.test.TestCase):
       sess.run(tf.global_variables_initializer())
       legacy_class_out, legacy_box_out = sess.run(feats)
     for i in range(3, 8):
-      self.assertAllClose(keras_class_out[i - 3],
-                          legacy_class_out[i],
-                          rtol=1e-4,
-                          atol=1e-4)
-      self.assertAllClose(keras_box_out[i - 3],
-                          legacy_box_out[i],
-                          rtol=1e-4,
-                          atol=1e-4)
+      self.assertAllClose(
+          keras_class_out[i - 3], legacy_class_out[i], rtol=1e-4, atol=1e-4)
+      self.assertAllClose(
+          keras_box_out[i - 3], legacy_box_out[i], rtol=1e-4, atol=1e-4)
 
     feats = tf.ones(inputs_shape)
     model = efficientdet_keras.EfficientDetNet(config=config)
     model.load_weights(tmp_ckpt)
     eager_class_out, eager_box_out = model(feats)
     for i in range(3, 8):
-      self.assertAllClose(eager_class_out[i - 3],
-                          legacy_class_out[i],
-                          rtol=1e-4,
-                          atol=1e-4)
-      self.assertAllClose(eager_box_out[i - 3],
-                          legacy_box_out[i],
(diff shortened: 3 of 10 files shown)


efficientdet/keras/train.py
```diff
@@ -0,0 +1,266 @@
...
+        use_fake_data=FLAGS.use_fake_data,
+        max_instances_per_image=config.max_instances_per_image)(params)
+
+  if strategy:
+    train_dataset = strategy.experimental_distribute_datasets_from_function(
+        functools.partial(get_dataset, is_training=True, params=params))
+    eval_dataset = strategy.experimental_distribute_datasets_from_function(
+        functools.partial(get_dataset, is_training=False, params=params))
+
+  # TODO(fsx950223): add train logic
```
mingxingtan, line 262: After you adding the train logic, I would be happy to test it on TPU.

efficientdet/keras/train.py
```diff
@@ -0,0 +1,266 @@
...
+  # Check data path
+  if FLAGS.mode in ('train',
+                    'train_and_eval') and FLAGS.training_file_pattern is None:
+    raise RuntimeError('You must specify --training_file_pattern for training.')
+  if FLAGS.mode in ('eval', 'train_and_eval'):
+    if FLAGS.validation_file_pattern is None:
+      raise RuntimeError('You must specify --validation_file_pattern '
+                         'for evaluation.')
+
+  # The following is for spatial partitioning. `features` has one tensor while
```
mingxingtan, line 171: This spatial partition code is optional: it is a legacy code, and I never use it for training.
Feel free to delete them if that helps reduce the coding complexity.

efficientdet/keras/train.py
fsx950223: Use custom loop or keras.model.fit?

mingxingtan: Either way is good to me. I guess custom loop is more flexible? WDYT?

efficientdet/keras/train.py
```diff
@@ -0,0 +1,266 @@
...
+                     help='Number of TPU cores for training')
+flags.DEFINE_bool('use_spatial_partition', False, 'Use spatial partition.')
+flags.DEFINE_integer('num_cores_per_replica',
+                     default=8,
+                     help='Number of TPU cores per'
+                     'replica when using spatial partition.')
+flags.DEFINE_multi_integer(
+    'input_partition_dims', [1, 4, 2, 1],
+    'A list that describes the partition dims for all the tensors.')
+flags.DEFINE_integer('train_batch_size', 64, 'training batch size')
```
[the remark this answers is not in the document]
mingxingtan: It is total batch size.

efficientdet/keras/train.py
fsx950223: https://www.tensorflow.org/guide/keras/customizing_what_happens_in_fit
I prefer custom loop too before I read above tutorial.

efficientdet/keras/train_lib.py
```diff
@@ -0,0 +1,348 @@
...
+    total_loss: an integer tensor representing total loss reducing from
+      class and box losses from all levels.
+    cls_loss: an integer tensor representing total class loss.
+    box_loss: an integer tensor representing total box regression loss.
+    box_iou_loss: an integer tensor representing total box iou loss.
+  """
+  # Sum all positives in a batch for normalization and avoid zero
+  # num_positives_sum, which would lead to inf loss during training
+  def __init__(self, params, **kwargs):
+    super(DetectionLoss, self).__init__(**kwargs)
```
fsx950223, line 275: We could use super() directly in py3, WDYT?

efficientdet/keras/postprocess.py
```diff
@@ -160,17 +160,17 @@ def nms(params, boxes: T, scores: T, classes: T, padded: bool) -> List[T]:
...
   # TF API's sigma is twice as the paper's value, so here we divide it by 2:
   # https://github.com/tensorflow/tensorflow/issues/40253.
-  nms_top_idx, nms_scores, nms_valid_lens = tf.raw_ops.NonMaxSuppressionV5(
+  nms_top_idx, nms_scores, _ = tf.raw_ops.NonMaxSuppressionV5(
       boxes=boxes,
       scores=scores,
       max_output_size=max_output_size,
       iou_threshold=iou_thresh,
-      score_threshold=score_thresh,
+      score_threshold=0,
```
fsx950223, line 168: We can set score_threshold=0, then filter results when we draw boxes by score.

mingxingtan: This might slow down the inference latency as NMS needs to process all boxes.
Actually I was thinking the opposite direction: if we filter out boxes based on scores much earlier (in pre_nms), then inference should be faster.
What do you think?

efficientdet/keras/train_lib.py
mingxingtan: Good point, no need to support python2 anymore. super() is good enough.

efficientdet/keras/utils_keras.py
```diff
@@ -45,7 +45,11 @@ def build_batch_norm(is_training_bn: bool,
     A normalized `Tensor` with the same `data_format`.
   """
   axis = 1 if data_format == 'channels_first' else -1
-  batch_norm_class = utils.batch_norm_class(is_training_bn, strategy)
+  if is_training_bn:
+    batch_norm_class = tf.keras.layers.experimental.SyncBatchNormalization
```
mingxingtan, line 49: Does SyncBatchNormalization work well on GPUs?

Could you add the following two lines:

```
# TODO(tanmingxing): compare them on TPU. 
# batch_norm_class = utils.batch_norm_class(is_training_bn, strategy)
```

efficientdet/keras/train.py
```diff
@@ -0,0 +1,211 @@
...
+          input_context, is_training=False, params=params))
+
+  with ds_strategy.scope():
+    model = EfficientDetNet(params['model_name'], config)
+    ckpt_path = tf.train.latest_checkpoint(FLAGS.model_dir)
+    if ckpt_path:
+      model.load_weights(ckpt_path)
+    model.compile(optimizer=get_optimizer(params), loss=[DetectionLoss(params)])
+  # TODO(fsx950223): override train_step and test_step
+  model.fit(train_dataset,
```
mingxingtan, line 201: Awesome. This training loop looks super simple!

efficientdet/keras/postprocess.py
fsx950223: I agree.

efficientdet/keras/utils_keras.py
fsx950223: I met the same [issue](https://github.com/tensorflow/tensorflow/issues/39569) before.
Could you test batch renormalization which appears better with small batch size?

fsx950223:
The training loop is finished. You could have a try.@mingxingtan

efficientdet/keras/postprocess.py
```diff
@@ -160,7 +162,7 @@ def nms(params, boxes: T, scores: T, classes: T, padded: bool) -> List[T]:
 
   # TF API's sigma is twice as the paper's value, so here we divide it by 2:
   # https://github.com/tensorflow/tensorflow/issues/40253.
-  nms_top_idx, nms_scores, nms_valid_lens = tf.raw_ops.NonMaxSuppressionV5(
```
mingxingtan, line 163: I guess we still need to maintain the nms_valid_llens since we do have some paddings? Otherwise, users may not know how many boxes are valid in the returned values.

efficientdet/det_model_fn.py
```diff
@@ -237,11 +237,6 @@ class and box losses from all levels.
   box_losses = []
   box_iou_losses = []
   for level in levels:
-    if params['data_format'] == 'channels_first':
```
mingxingtan, line 240: So is this a bug in previous code?

efficientdet/keras/postprocess.py
fsx950223: As scores are padded by 0, so users could filter it by score again.

efficientdet/det_model_fn.py
fsx950223: I move it to dataset.

efficientdet/dataloader.py
```diff
@@ -382,7 +382,15 @@ def _process_example(images, cls_targets, box_targets, num_positives,
               batch_size,
           ]), [batch_size, 1])
 
+      if params['data_format'] == 'channels_first':
```
fsx950223, line 385: Here.

efficientdet/keras/postprocess.py
fsx950223: Maybe we could output a ragged tensor. It's interesting.

fsx950223:
Current implementation still has lots of bugs.

efficientdet/keras/train_lib.py
```diff
@@ -0,0 +1,477 @@
...
+  @tf.function
+  def test_step(self, features, labels):
+    cls_outputs, box_outputs = self(features, training=False)
+    det_loss, cls_loss, box_loss, box_iou_loss = self._detection_loss(
+        cls_outputs, box_outputs, labels)
+    reg_l2loss = self._reg_l2_loss(self.config.weight_decay)
+    total_loss = det_loss + reg_l2loss
+    return total_loss
+
+  def train(self,
```
mingxingtan, line 450: Should we use model.fit/evaluate according to https://www.tensorflow.org/guide/keras/customizing_what_happens_in_fit?

efficientdet/keras/train.py
```diff
@@ -0,0 +1,220 @@
...
+  with ds_strategy.scope():
+    model = EfficientDetNetTrain(params['model_name'], config)
+  ckpt_path = tf.train.latest_checkpoint(FLAGS.model_dir)
+  if ckpt_path:
+    model.load_weights(ckpt_path)
+  model.freeze_vars(params['var_freeze_expr'])
+  with ds_strategy.scope():
+    model.compile(
+        optimizer=get_optimizer(params),
+        box_loss=BoxLoss(params['delta'],
```
mingxingtan, line 205: I don't quite understand this. Should compile just take two parameters: optimizer and loss?  (the loss can be a dict or a string or a losses instance)?

Could you add the following test? (I tried it, but it failed)

  def test_train(self):
    config = hparams_config.get_detection_config('efficientdet-d0')
    config.batch_size = 1
    config.num_examples_per_epoch = 1
    x = np.random.random((1, 512, 512, 3))
    labels = {
        'box_targets_%d' % i: np.ones((1, 512 // 2**i, 512 // 2**i, 810))
        for i in range(3, 8)
    }
    labels.update({
        'cls_targets_%d' % i: np.random.random(
            (1, 512 // 2**i, 512 // 2**2, 36)) for i in range(3, 8)
    })
    labels.update({'mean_num_positives': 100})

    model = train_lib.EfficientDetNetTrain(config=config)
    params = config.as_dict()
    model.compile(
        optimizer=train_lib.get_optimizer(params),
    )
    outputs = model.train_on_batch(x, labels)

fsx950223: I have overrided compile and implement train loop by myself, so you couldn't use train_on_batch and have to specific box_loss, box_iou_loss, class_loss

fsx950223: I try to transfer to train_on_batch

mingxingtan: Aha, I see. What's the benefits to override compile (looks like it just simplifies the loss)? Should we combine those loss functions into a single loss, and then pass it to compile?

fsx950223: I have migrated to train_on_batch

fsx950223:
https://github.com/google/automl/blob/14548b7175e093c9f0586b372180c41ffc04fbc1/efficientdet/keras/postprocess.py#L253
why padded is False.

mingxingtan:
> https://github.com/google/automl/blob/14548b7175e093c9f0586b372180c41ffc04fbc1/efficientdet/keras/postprocess.py#L253
> 
> 
> why padded is False.

Per_class nms is very special: you don't want to pad each class because a lot of classes would be empty for a single image, so  set padded to False for each class. Because there is no padd for each class, we need to pad the final concat results. 

Another alternative is to pad each class, then you will have 90 * MAX_OUTPUT_SIZE outputs, then you can do another topK, but I believe it would make the inference unnecessary slow.

fsx950223:
Yeah, 90*100*math.log(100) is a little smaller than 9000*math.log(9000).

efficientdet/keras/train_lib_test.py
```diff
@@ -0,0 +1,85 @@
...
+    labels.update({
+        'cls_targets_%d' % i: tf.ones((1, 512 // 2**i, 512 // 2**i, 9),
+                                      dtype=tf.int32) for i in range(3, 8)
+    })
+    labels.update({'mean_num_positives': 100.0})
+
+    model = EfficientDetNetTrain(config=config)
+    params = config.as_dict()
+    model.compile(optimizer=get_optimizer(params),
+                  loss={
```
mingxingtan, line 66: I still don't understand how loss is computed here.  In the keras doc, it says it would compute all losses in this dict and sum them up.

Here each loss takes some tensors as inputs, but the network outputs and groundtruth are a list of tensors for different levels. How they are connected?

fsx950223: I use same logic as origin code.

mingxingtan: So this loss parameter here is not really useful? (the doc says if you pass in loss, then you can use something like compiled_loss)

fsx950223: I don't use self.compiled_loss, I use self.loss instead. It's a dict.

fsx950223: I can't monitor box loss and class loss if I use compiled_loss.

mingxingtan: Gotcha!

efficientdet/keras/train.py
```diff
@@ -0,0 +1,214 @@
...
+                      "box_iou_loss":
+                          BoxIouLoss(params['iou_loss_type'],
+                                     reduction=tf.keras.losses.Reduction.NONE),
+                      "class_loss":
+                          FocalLoss(params['alpha'],
+                                    params['gamma'],
+                                    label_smoothing=params['label_smoothing'],
+                                    reduction=tf.keras.losses.Reduction.NONE)
+                  })
+    model.train(get_callbacks(params, FLAGS.profile),
```
mingxingtan, line 207: The keras doc recommends  "model.fit" instead of train. Should we use model.fit?

fsx950223: ```model.fit``` process input by nest.flatten and zip, which only return top10 values in labels dict.

mingxingtan: Maybe we should convert the labels into a list as well?
I roughly read the fit function: looks like it does a lot of extra work for multi-worker and multi-GPU/TPU training. It might be difficult to capture all of them in the customized train. WDYT?

fsx950223: It could be done in next PR.
In fact, I have used the custom train loop in my repo for half a year.

mingxingtan: Aha, okay, then let's first try to use this customized train.

efficientdet/keras/train_lib.py
```diff
@@ -0,0 +1,474 @@
...
+    total_loss = (cls_loss + self.config.box_loss_weight * box_loss +
+                  self.config.iou_loss_weight * box_iou_loss)
+    return total_loss, cls_loss, box_loss, box_iou_loss
+
+  def _steps_loop(self, steps, dataset_iter, is_train, logs, callbacks):
+    total_loss = 0.0
+    for step in (1, steps + 1):
+      self.step.assign(step)
+      features, labels = next(dataset_iter)
+      [callback.on_batch_begin(step, logs) for callback in callbacks]
```
mingxingtan, line 384: list comprehension looks a little strange here because the list is not used anywhere (it will also triggler warning "expression-not-assigned".

How about change it to:

`map(lambda x, s=step: x.on_batch_begin(s, logs), callbacks)`

efficientdet/keras/train_lib.py
mingxingtan: As discussed in the other comment, let's just use train in this PR.

Could you help add some doc or link for this train function? (looks like it is not a member of tf.keras.Model?)  
Also, it would be great if we can somehow add a simple test to train_lib_test.py

efficientdet/keras/train_lib_test.py
```diff
@@ -0,0 +1,85 @@
...
+                      "box_iou_loss":
+                          BoxIouLoss(params['iou_loss_type'],
+                                     reduction=tf.keras.losses.Reduction.NONE),
+                      "class_loss":
+                          FocalLoss(params['alpha'],
+                                    params['gamma'],
+                                    label_smoothing=params['label_smoothing'],
+                                    reduction=tf.keras.losses.Reduction.NONE)
+                  })
+    outputs = model.train_on_batch(x, labels)
```
mingxingtan, line 79: If you print out outputs, it is empty. Shouldn't this return a dict {'loss': total_loss}?

mingxingtan: I debugged a little bit. This is because return_dict=False in default, but we did not set metrics during compile.
Change it to `outputs = model.train_on_batch(x, labels, return_dict=True)` will return a dict {'loss': loss_v}

fsx950223:
I modify them later.

efficientdet/keras/train_lib_test.py
fsx950223: Something wrong with my memory. I have migrated to ```model.fit```.

efficientdet/keras/train.py
```diff
@@ -0,0 +1,210 @@
...
+                mode=FLAGS.mode)
+
+  # set mixed precision policy by keras api which will generate mixed precision model
+  precision = utils.get_precision(params['strategy'], params['mixed_precision'])
+  policy = tf.keras.mixed_precision.experimental.Policy(precision)
+  tf.keras.mixed_precision.experimental.set_policy(policy)
+
+  def get_dataset(is_training, params):
+    global_batch_size = FLAGS.batch_size
+    batch_size = global_batch_size // ds_strategy.num_replicas_in_sync
```
mingxingtan, line 172: Where is this batch_size used? 
Looks like it should be params['batch_size'] = ...?

efficientdet/keras/train_lib_test.py
mingxingtan: fit seems to be cleaner!

efficientdet/keras/efficientdet_keras_test.py
```diff
@@ -66,20 +66,28 @@ def test_model_output(self):
...
     feats = tf.ones(inputs_shape)
     model = efficientdet_keras.EfficientDetNet(config=config)
     model.load_weights(tmp_ckpt)
     eager_class_out, eager_box_out = model(feats)
     for i in range(3, 8):
-      self.assertAllClose(
-        eager_class_out[i - 3], legacy_class_out[i], rtol=1e-4, atol=1e-4)
-      self.assertAllClose(
-        eager_box_out[i - 3], legacy_box_out[i], rtol=1e-4, atol=1e-4)
+      self.assertAllClose(eager_class_out[i - 3],
```
mingxingtan, line 83: I realized our formater difference is caused by the yapf version difference.

You probably use the stable yapf-0.30, but the internal version is based on the latest yapf. They are mostly the same with a few small differences.

Could you add an option "SPLIT_BEFORE_EXPRESSION_AFTER_OPENING_PAREN = True" in your yapf config and rerun your auto-format for this PR? (same as the latest yapf code: https://github.com/google/yapf/blob/51ffe2d07930a509ecb2ef454a7b251eeeac0a59/yapf/yapflib/style.py#L502)

fsx950223:
I have reformated the code.

PR #538, opened 2020-06-23
Merge the keras layer tests.

mingxingtan:


efficientdet/keras/efficientdet_arch_keras_test.py
```diff
@@ -100,31 +100,27 @@ def test_build_feature_network(self):
...
 
   def test_model_variables(self):
-    feats = tf.ones([1, 512, 512, 3])
-    model = efficientdet_arch_keras.efficientdet('efficientdet-d0')
-    model(feats)
-    vars1 = [var.name for var in model.trainable_variables]
-    vars2 = [var.name for var in model.variables]
+    # TODO(tanmingxing): Re-enable this code once pass internal tests.
+    # feats = tf.ones([1, 512, 512, 3])
+    # model = efficientdet_arch_keras.efficientdet('efficientdet-d0')
```
mingxingtan, line 105: This code passes with public tennsorflow release, but fails with internal code (ahead of public release). Error is:

google3.third_party.tensorflow.python.framework.errors_impl.FailedPreconditionError:  Error while reading resource variable _AnonymousVar525 from Container: localhost. This could mean that the variable was uninitialized. Not found: Resource localhost/_AnonymousVar525/N10tensorflow3VarE does not exist.
	 [[node fpn_cells/cell_2/fnode3/op_after_combine8/bn/ReadVariableOp (defined at /third_party/brain_automl/efficientdet/keras/efficientdet_arch_keras.py:715) ]] [Op:__inference_keras_scratch_graph_8420]

So I temporarily disable it for now

fsx950223: Could it run in eager mode out of keras test.

mingxingtan: Interesting. If I run these code in a binary file, it works well.  It only fails in testcase.
Do you have any idea on this?

fsx950223: > Interesting. If I run these code in a binary file, it works well. It only fails in testcase.
> Do you have any idea on this?

Maybe you should init vars when run test code with interial tensorflow. I guess internal tensorflow use graph mode as default.

Issue #539, opened 2020-06-23
load a trained weight 


Samjith888:
In the repo, model_inspect.py files is using for inference and model conversion and related stuffs. Will u make script to load a trained model(.pb file) for inference. 

Eg:

```
import tensorflow as tf
from efficientdet import load_model

effdet = load_model('trained_model.pb')
im_size = model.config.input_size
images  = tf.random.uniform((3, *im_size, 3)) # 3 Mock images

boxes, labels, scores = effdet(images, training=False)

for im_boxes, im_labels in zip(boxes, labels):
    # Process boxes of a specific image
```

@mingxingtan @fsx950223

mingxingtan:
Hi @Samjith888 
If you specify --saved_model_dir as a pb file, then it would use the pb file to do inference, like this:

```
!python model_inspect.py --runmode=saved_model_infer \
  --model_name=efficientdet-d0  \
  --saved_model_dir=trained_model.pb  \
  --input_image=img.png --output_image_dir=/tmp/
```

Samjith888:
> Hi @Samjith888 
> 
> If you specify --saved_model_dir as a pb file, then it would use the pb file to do inference, like this:
> 
> 
> 
> ```
> 
> !python model_inspect.py --runmode=saved_model_infer \
> 
>   --model_name=efficientdet-d0  \
> 
>   --saved_model_dir=trained_model.pb  \
> 
>   --input_image=img.png --output_image_dir=/tmp/
> 
> ```

 I meant can u just  make one line code to load the model

PR #542, opened 2020-06-24
Use keras models for inference.

mingxingtan:


efficientdet/inference.py
```diff
@@ -161,11 +161,19 @@ def build_model(model_name: Text, inputs: tf.Tensor, **kwargs):
     (cls_outputs, box_outputs): the outputs for class and box predictions.
     Each is a dictionary with key as feature level and value as predictions.
   """
-  model_arch = det_model_fn.get_model_arch(model_name)
   mixed_precision = kwargs.get('mixed_precision', None)
   precision = utils.get_precision(kwargs.get('strategy', None), mixed_precision)
-  cls_outputs, box_outputs = utils.build_model_with_precision(
-      precision, model_arch, inputs, False, model_name, **kwargs)
+
+  if kwargs.get('use_keras_model', None):
```
mingxingtan, line 167: @fsx950223 @Ely-S @VicariNishimura 

Hi, I am trying to use keras model as default for inference, but it seems like it doesn't produce the expected results.

```
python model_inspect.py --runmode=saved_model --model_name=efficientdet-d0 \
  --hparams='image_size=1920x1280,use_keras_model=true' \
  --ckpt_path=efficientdet-d0  --saved_model_dir=/tmp/savedmodel

python model_inspect.py --runmode=saved_model_infer --model_name=efficientdet-d0 \
  --input_image=img.png --output_image_dir=/tmp/imgout/ \
  --saved_model_dir=/tmp/savedmodel/
```
It successfully loads the checkpoint and generate an output image, but the output image has very few boxes like this:



Do you have any idea what's the problem here? Thanks!

fsx950223: It seems something wrong with image_size.

mingxingtan: I also thought that's about image size, then I tried  --hparams='use_keras_model=true' and got zero box. So I guess there might be some other issues.

efficientdet/keras/efficientdet_arch_keras_test.py
```diff
@@ -58,15 +58,24 @@ def test_model_output(self):
...
-      feats = legacy_arch.build_feature_network(feats, config)
-      feats = legacy_arch.build_class_and_box_outputs(feats, config)
+      feats = legacy_arch.efficientdet(feats, config=config)
       sess.run(tf.global_variables_initializer())
       class_output2, box_output2 = sess.run(feats)
     for i in range(3, 8):
       self.assertAllEqual(class_output1[i], class_output2[i])
       self.assertAllEqual(box_output1[i], box_output2[i])
 
+    with tf.Session(graph=tf.Graph()) as sess:
```
mingxingtan, line 68: @fsx950223 

I added a failing test. It turns out 

```
      feats, _ = efficientdet_arch_keras.build_backbone(feats, config)
      feats = efficientdet_arch_keras.build_feature_network(feats, config)
      feats = efficientdet_arch_keras.build_class_and_box_outputs(feats, config)
```

doesn't equal to:

```
      feats, _ = efficientdet_arch_keras.efficientdet(config)(feats)
```

I am surprised by this, because in efficientdet_arch_keras.efficientdet, it essential just calls the three function. The only difference is the "inputs = tf.keras.layers.Input([*utils.parse_image_size(config.image_size), 3])".

Any thoughts?

fsx950223: It's strange, I'm diving into it.

efficientdet/inference.py
fsx950223: > I also thought that's about image size, then I tried --hparams='use_keras_model=true' and got zero box. So I guess there might be some other issues.

I fix the bug by flowing changes:
```python
    def model_arch(feats, model_name=None, config=None, **kwargs):
      if not config:
        config = hparams_config.get_efficientdet_config(model_name)
      elif isinstance(config, dict):
        config = hparams_config.Config(config)  # wrap dict in Config object
      if kwargs:
        config.override(kwargs)
      model = efficientdet_arch_keras.efficientdet(feats, model_name, config=config)
      cls_outputs, box_outputs = model.outputs[1:6], model.outputs[6:]
      cls_outputs = {index+3:output for index, output in enumerate(cls_outputs)}
      box_outputs = {index+3:output for index, output in enumerate(box_outputs)}
      return cls_outputs, box_outputs
    cls_outputs, box_outputs = utils.build_model_with_precision(
        precision, model_arch, inputs, False, model_name, **kwargs)
```
```python
def efficientdet(features, model_name=None, config=None, **kwargs):
  inputs = tf.keras.layers.Input(
      [*utils.parse_image_size(config.image_size), 3], tensor=features)
```

fsx950223: > > I also thought that's about image size, then I tried --hparams='use_keras_model=true' and got zero box. So I guess there might be some other issues.
> 
> I fix the bug by following changes:
> 
> ```python
>     def model_arch(feats, model_name=None, config=None, **kwargs):
>       if not config:
>         config = hparams_config.get_efficientdet_config(model_name)
>       elif isinstance(config, dict):
>         config = hparams_config.Config(config)  # wrap dict in Config object
>       if kwargs:
>         config.override(kwargs)
>       model = efficientdet_arch_keras.efficientdet(feats, model_name, config=config)
>       cls_outputs, box_outputs = model.outputs[1:6], model.outputs[6:]
>       cls_outputs = {index+3:output for index, output in enumerate(cls_outputs)}
>       box_outputs = {index+3:output for index, output in enumerate(box_outputs)}
>       return cls_outputs, box_outputs
>     cls_outputs, box_outputs = utils.build_model_with_precision(
>         precision, model_arch, inputs, False, model_name, **kwargs)
> ```
> 
> ```python
> def efficientdet(features, model_name=None, config=None, **kwargs):
>   inputs = tf.keras.layers.Input(
>       [*utils.parse_image_size(config.image_size), 3], tensor=features)
> ```

The strange behavior is caused by placeholder.

mingxingtan: @fsx950223 I checked in a commit with your fix, but somehow still couldn't get the expected output. I might have missed something.
Could you take a look or check in your change to this branch (you should have write access now)? Thanks!

fsx950223:
https://github.com/fsx950223/automl/tree/keras_fix
Could you test model by infer mode?

mingxingtan:
> https://github.com/fsx950223/automl/tree/keras_fix
> Could you test model via infer mode?

Excellent, it works well now!  The trick is to use model.outputs instead of calling the model to get the outputs. Thank you, @fsx950223 
We can go with this for now, but I think we should provide a cleaner way to use keras model for inference. something like #539 asked.

fsx950223:
> > https://github.com/fsx950223/automl/tree/keras_fix
> > Could you test model via infer mode?
> 
> Excellent, it works well now! The trick is to use model.outputs instead of calling the model to get the outputs. Thank you, @fsx950223
> We can go with this for now, but I think we should provide a cleaner way to use keras model for inference. something like #539 asked.

We can simplify ServingDriver usage.
```python
serve=ServingDriver()
serve.load(saved_model)
```

Samjith888:
Please add a section like 'Load a trained weight' in README.md

PR #549, opened 2020-06-26
Ensure same layer names under eager and graph mode.

mingxingtan:
PiperOrigin-RevId: 318538945

### DIFF
--- a/efficientdet/backbone/efficientnet_model.py
+++ b/efficientdet/backbone/efficientnet_model.py
@@ -25,6 +25,7 @@ from __future__ import print_function
 
 import collections
 import functools
+import itertools
 import math
 
 from absl import logging
@@ -211,6 +212,17 @@ class MBConvBlock(tf.keras.layers.Layer):
 
   def _build(self):
     """Builds block according to the arguments."""
+    # pylint: disable=g-long-lambda
+    # Keras layers have different names under graph and eager mode, so here we
+    # force their names to be the same for compatible.
+    bn_id = itertools.count(0)
+    get_bn_name = lambda: 'tpu_batch_normalization' if not next(
+        bn_id) else 'tpu_batch_normalization_' + str(next(bn_id) // 2)
+    conv_id = itertools.count(0)
+    get_conv_name = lambda: 'conv2d' if not next(
+        conv_id) else 'conv2d_' + str(next(conv_id) // 2)
+    # pylint: enable=g-long-lambda
+
     if self._block_args.super_pixel == 1:
       self._superpixel = self.conv_cls(
           self._block_args.input_filters,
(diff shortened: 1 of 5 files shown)


efficientdet/keras/efficientdet_arch_keras_test.py
```diff
@@ -51,31 +51,32 @@ def test_model_output(self):
...
       feats = tf.ones(inputs_shape)
       tf.random.set_random_seed(SEED)
       feats = legacy_arch.efficientdet(feats, config=config)
       sess.run(tf.global_variables_initializer())
       legacy_class_out, legacy_box_out = sess.run(feats)
     for i in range(3, 8):
-      self.assertAllEqual(keras_class_out[i - 3], legacy_class_out[i])
-      self.assertAllEqual(keras_box_out[i - 3], legacy_box_out[i])
+      self.assertAllClose(
+          keras_class_out[i - 3], legacy_class_out[i], rtol=1e-4, atol=1e-4)
```
mingxingtan, line 68: @fsx950223 
With rtol and ato,  EfficientDetModel can close results in graph mode, but the difference in eager mode is still big (line 78/79).

efficientdet/keras/efficientdet_arch_keras.py
```diff
@@ -812,7 +812,7 @@ def __init__(self, model_name=None, config=None, name=''):
     # Feature network.
     feat_sizes = utils.get_feat_sizes(config.image_size, config.max_level)
     self.resample_layers = {}  # additional resampling layers.
-    for level in range(config.min_level, config.max_level + 1):
+    for level in range(6, config.max_level + 1):
```
fsx950223, line 815: Why?

fsx950223:
```python
  def test_model_output(self):
    inputs_shape = (1, 512, 512, 3)
    config = hparams_config.get_efficientdet_config('efficientdet-d0')
    ckpt_path = '/tmp/test_ckpt'

    with tf.Session(graph=tf.Graph()) as sess:
      feats = tf.ones(inputs_shape)
      feats = legacy_arch.efficientdet(feats, config=config)
      sess.run(tf.global_variables_initializer())
      legacy_class_out, legacy_box_out = sess.run(feats)
      all_saver = tf.train.Saver()
      all_saver.save(sess, ckpt_path)

    ckpt = tf.train.load_checkpoint(ckpt_path)
    with tf.Session(graph=tf.Graph()) as sess:
      feats = tf.ones(inputs_shape)
      feats = efficientdet_arch_keras.build_backbone(feats, config)
      feats = efficientdet_arch_keras.build_feature_network(feats, config)
      feats = efficientdet_arch_keras.build_class_and_box_outputs(feats, config)
      variables = tf.global_variables()
      assign_ops = [var.assign(ckpt.get_tensor(var.name.split(':')[0])) for var in variables]

      sess.run(assign_ops)
      keras_class_out, keras_box_out = sess.run(feats)
    for i in range(3, 8):
      self.assertAllClose(keras_class_out[i - 3], legacy_class_out[i], rtol=1e-4, atol=1e-4)
      self.assertAllClose(keras_box_out[i - 3], legacy_box_out[i], rtol=1e-4, atol=1e-4)
 
```
if the weights are loaded from a checkpoint, then the outputs are different. But if the weights are initialized from random values (with the same random seed), then the outputs are same.
Do you have any idea about this?

fsx950223:
The current implementation has some bug on names?

efficientdet/backbone/efficientnet_model.py
```diff
@@ -211,6 +212,17 @@ def block_args(self):
...
+    # pylint: disable=g-long-lambda
+    # Keras layers have different names under graph and eager mode, so here we
+    # force their names to be the same for compatible.
+    bn_id = itertools.count(0)
+    get_bn_name = lambda: 'tpu_batch_normalization' if not next(
+        bn_id) else 'tpu_batch_normalization_' + str(next(bn_id) // 2)
+    conv_id = itertools.count(0)
+    get_conv_name = lambda: 'conv2d' if not next(
+        conv_id) else 'conv2d_' + str(next(conv_id) // 2)
+    # pylint: enable=g-long-lambda
```
fsx950223, line 224: Keras does this by default in eager and graph mode.

efficientdet/keras/efficientdet_arch_keras.py
mingxingtan: I am trying to use model.summry(), but it reports errors if some variables are not used in call function.
Do you  know some better solutions?

efficientdet/backbone/efficientnet_model.py
mingxingtan: Yeah, this could be a potentially internal bug (or future TF feature, I am not sure, need to double confirm with TF team). We have discussed this in https://github.com/google/automl/pull/538 (It cannot pass internal tests, but it runs well in binary environment).

Since the name changes are relatively simple, we can keep this for now. If it is indeed an internal bug, then we could remove this code after fixing that bug.

efficientdet/backbone/efficientnet_model.py
```diff
@@ -219,53 +231,61 @@ def _build(self):
...
-    self._depthwise_conv = self.depthwise_conv_cls(
-        kernel_size=[kernel_size, kernel_size],
-        strides=self._block_args.strides,
-        depthwise_initializer=conv_kernel_initializer,
-        padding='same',
-        data_format=self._data_format,
-        use_bias=False)
+    if self._block_args.fused_conv:
+      # Fused expansion phase. Called if using fused convolutions.
+      self._fused_conv = self.conv_cls(
```
mingxingtan, line 247: @fsx950223 these changes are also for supporting model.summary.

mingxingtan: github did a bad job on diff (the actual changes are pretty simple: added if, else, and name=).

fsx950223:
I suggest using the release tf version to develop the model. Canary and Nightly builds may have some bugs and experimental features that will be removed soon.

efficientdet/backbone/efficientnet_model.py
fsx950223: I don't recommend use summary with subclass model.
It's design for keras graph model.

efficientdet/keras/efficientdet_arch_keras.py
fsx950223: I agree with this change.
It should be commented on how to get the value.

VicariNishimura:
LGTM

PR #569, opened 2020-07-07
Clean up backbone model to avoid name scope.

mingxingtan:


efficientdet/keras/utils_keras.py
```diff
@@ -48,7 +48,7 @@ def build_batch_norm(is_training_bn: bool,
   """
   axis = 1 if data_format == 'channels_first' else -1
   if is_training_bn:
-    if strategy in ('gpus', None):
+    if strategy in ('gpus',):
```
mingxingtan, line 51: This change is needed to pass the efficientdet_keras_test.py  @fsx950223

fsx950223:
```
strategy == 'gpus' and is_training_bn
```
```
strategy == 'tpu' and is_training_bn
```
to enable fused batch norm on single gpu.

mingxingtan: Sorry, what do you mean by "strategy == 'tpu' and is_training_bn"?

fsx950223: > Sorry, what do you mean by "strategy == 'tpu' and is_training_bn"?

```python
  if strategy == 'gpus' and is_training_bn:
    batch_norm_class = tf.keras.layers.experimental.SyncBatchNormalization
  elif strategy == 'tpu' and is_training_bn:
    # TODO(tanmingxing): compare them on TPU.
    batch_norm_class = utils.batch_norm_class(is_training_bn, strategy)
  else:
    batch_norm_class = tf.keras.layers.BatchNormalization
```

fsx950223: No need to change it, I misunderstand

fsx950223:
LGTM

PR #573, opened 2020-07-08
add logs to tensorboard

fsx950223:
For tensorboard vis

### DIFF
--- a/efficientdet/keras/train_lib.py
+++ b/efficientdet/keras/train_lib.py
@@ -17,11 +17,14 @@
 import math
 import re
 from absl import logging
+import numpy as np
 import tensorflow as tf
 
 import iou_utils
+import utils
+import inference
 from keras import efficientdet_keras
-
+from concurrent.futures import ThreadPoolExecutor
 
 def update_learning_rate_schedule_parameters(params):
   """Updates params that are related to the learning rate schedule."""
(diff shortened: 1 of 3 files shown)


efficientdet/keras/train_lib.py
```diff
@@ -427,7 +427,8 @@ def train_step(self, data):
     images, labels = data
     with tf.GradientTape() as tape:
       cls_outputs, box_outputs = self(images, training=True)
-      det_loss, _, _, _ = self._detection_loss(cls_outputs, box_outputs, labels)
+      det_loss, cls_loss, box_loss, box_iou_loss \
```
mingxingtan, line 430: Prefer 

```
det_loss, cls_loss, box_loss, box_iou_loss = (
    self._detection_loss(cls_outputs, box_outputs, labels))
```

efficientdet/keras/train_lib.py
```diff
@@ -467,7 +473,12 @@ def test_step(self, data):
     """
     images, labels = data
     cls_outputs, box_outputs = self(images, training=False)
-    det_loss, _, _, _ = self._detection_loss(cls_outputs, box_outputs, labels)
+    det_loss, cls_loss, box_loss, box_iou_loss =\
```
mingxingtan, line 476: Similar change (better to avoid \)

efficientdet/hparams_config.py
```diff
@@ -249,6 +249,8 @@ def default_detection_configs():
 
   # For post-processing nms, must be a dict.
   h.nms_configs = {'method': 'hard'}
+  h.nms_score_thresh = 0.3
```
mingxingtan, line 252: How about put all these to nms_configs (postprocess.py is assuming these):

```
h.nms_configs = {
  'method': 'hard',
  'score_thresh': 0.5,
  'nms_max_output_size: 100,
}
```

I am also consider making this config file hierarchical like:

```
train:
  xxx

eval:
  yyy

nms:
  zzz
```

What do you think?

fsx950223: Good point.

fsx950223:
Is it a bug?
https://github.com/google/automl/blob/88b9a478cee2c4d79738af79d443475b79c19d45/efficientdet/iou_utils.py#L177
pred_boxes should be [ymin, xmin, ymax, xmax] instead of box regression.
And https://github.com/google/automl/blob/de4bdd67211fd36f54c7b19ae5a2a87618ad48e2/efficientdet/dataloader.py#L271 seems wrong.

efficientdet/keras/train_lib.py
mingxingtan: > Is it a bug?
> https://github.com/google/automl/blob/88b9a478cee2c4d79738af79d443475b79c19d45/efficientdet/iou_utils.py#L177
> 
> 
> pred_boxes should be [ymin, xmin, ymax, xmax] instead of box regression which is between [0,1].]]

Hmm, I haven't looked closely to iou_loss_utils.py. but it seems like " mask - tf.squeeze(_iou_per_anchor(pred_boxes, target_boxes, iou_type)))" should be " mask * (1 - tf.squeeze(_iou_per_anchor(pred_boxes, target_boxes, iou_type))))"

> And comments
> https://github.com/google/automl/blob/de4bdd67211fd36f54c7b19ae5a2a87618ad48e2/efficientdet/dataloader.py#L271
> 
> seems wrong.

Not sure about this.

PR #574, opened 2020-07-08
Replace pyfunc postprocessing with the new keras utility.

mingxingtan:
Tested with d0 eval accuracy.

### DIFF
--- a/efficientdet/det_model_fn.py
+++ b/efficientdet/det_model_fn.py
@@ -31,6 +31,7 @@ import hparams_config
 import iou_utils
 import retinanet_arch
 import utils
+from keras import postprocess
 
 _DEFAULT_BATCH_SIZE = 64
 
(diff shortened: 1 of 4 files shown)


efficientdet/det_model_fn.py
```diff
@@ -582,14 +535,21 @@ def metric_fn(**kwargs):
         tf.tile(tf.expand_dims(box_loss, 0), [
             params['batch_size'],
         ]), [params['batch_size'], 1])
+
+    params['nms_configs']['max_nms_inputs'] = anchors.MAX_DETECTION_POINTS
+    detections_bs = postprocess.generate_detections(params, cls_outputs,
```
mingxingtan, line 540: @fsx950223 FYI, with this utility, eval doesn't depends on pyfunc anymore.

You may leverage this utility for keras train/eval.

fsx950223: LGTM

PR #577, opened 2020-07-08
fix importing distributions when tfp not available

drwaltman:
Summary:
Updated imports to fix accessing `Beta` and `Dirichlet` distributions when `tensorflow_probability` is not available for import.

Issue:
If `tensorflow_probability` is not available for import, `tensorflow.distributions` is mapped to `tfp` instead.  `tfp` is used for `tfp.distributions.Beta` and `tfp.distributions.Dirichlet` which provide the following error when called:
```
AttributeError: module 'tensorflow.distributions' has no attribute 'distributions'
```


drwaltman:
@googlebot I signed it!

PR #581, opened 2020-07-10
Add keras model eval support

mingxingtan:


fsx950223:
How about wrapper it as Keras callback?

efficientdet/coco_metric.py
```diff
@@ -96,130 +213,17 @@ def estimator_metric_fn(self, detections, groundtruth_data):
...
-      self.category_ids = list(set(self.category_ids))
-      self.dataset['categories'] = [
-          {'id': int(category_id)} for category_id in self.category_ids
-      ]
-
     with tf.name_scope('coco_metric'):
       if self.testdev_dir:
-        update_op = tf.py_func(_update_op, [detections, groundtruth_data], [])
-        metrics = tf.py_func(_evaluate, [], tf.float32)
+        update_op = tf.py_func(self.update_state,
```
fsx950223, line 218: Use tf.numpy_funciton or tf.py_function.

fsx950223:
https://github.com/google/automl/blob/c9faec9d5ebe5678c8f9e255e2d73e36874e3974/efficientdet/keras/train_lib.py#L438
Could I calculate each level loss together by stack them to a tensor, which may decrease memory usage.

mingxingtan:
> https://github.com/google/automl/blob/c9faec9d5ebe5678c8f9e255e2d73e36874e3974/efficientdet/keras/train_lib.py#L438
> 
> 
> Could I calculate each level loss together by stack them to a tensor, which may decrease memory usage.

Yes you can, but I guess the memory impact might be limited?

mingxingtan:
> How about wrapper it as Keras callback?

Good idea. Could you help wrap them and then use them in train.py?

mingxingtan:
@fsx950223 If this PR looks good to you, maybe I can merge it first, so you can wrap it as a callback later?

fsx950223:
> > How about wrapper it as Keras callback?
> 
> Good idea. Could you help wrap them and then use them in train.py?

I try it later.

PR #588, opened 2020-07-10
Add flags to the keras eval script.

LucasSloan:
Currently it hardcodes the location of the COCO dataset and the model to use.

### DIFF
--- a/efficientdet/keras/eval.py
+++ b/efficientdet/keras/eval.py
@@ -15,6 +15,7 @@
 # ==============================================================================
 """Eval libraries."""
 from absl import app
+from absl import flags
 from absl import logging
 import tensorflow as tf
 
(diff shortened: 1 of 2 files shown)


efficientdet/keras/eval.py
```diff
@@ -26,16 +27,36 @@
...
+    'COCO validation JSON containing golden bounding boxes.')
+flags.DEFINE_string('model_name', 'efficientdet-d0',
+                    'Model name: the efficientdet model to use.')
+flags.DEFINE_string('checkpoint', None, 'Location of the checkpoint to evaluate.')
+
+FLAGS = flags.FLAGS
+
 def main(_):
-  config = hparams_config.get_efficientdet_config('efficientdet-d0')
+  if FLAGS.validation_file_pattern is None:
```
mingxingtan, line 42: You can use flags.mark_flag_as_required('validation_file_pattern')
see: https://abseil.io/docs/python/guides/flags

mingxingtan:
Thanks Lucas. Would be great if you can also help update the infer.py file.

LucasSloan:
Would you like me to update infer.py in this PR?

mingxingtan:
Yep, the change should be very similar, no?

efficientdet/keras/eval.py
LucasSloan: Done.

efficientdet/keras/eval.py
```diff
@@ -15,36 +15,53 @@
...
+flags.DEFINE_string('validation_file_pattern', None,
+                    'Glob for evaluation tfrecords (e.g., COCO val2017 set)')
+flags.DEFINE_string(
+    'val_json_file', None,
+    'COCO validation JSON containing golden bounding boxes.')
+flags.DEFINE_string('checkpoint', None, 'Location of the checkpoint to evaluate.')
+flags.DEFINE_string('model_name', 'efficientdet-d0',
+                    'Model name: the efficientdet model to use.')
+
+flags.mark_flag_as_required('validation_file_pattern')
```
mingxingtan, line 40: You need to move this right after "if __name__ == '__main__'"

efficientdet/keras/eval.py
```diff
@@ -15,36 +15,53 @@
...
 from keras import anchors
 from keras import efficientdet_keras
 from keras import postprocess
 
+flags.DEFINE_string('validation_file_pattern', None,
+                    'Glob for evaluation tfrecords (e.g., COCO val2017 set)')
+flags.DEFINE_string(
+    'val_json_file', None,
+    'COCO validation JSON containing golden bounding boxes.')
+flags.DEFINE_string('checkpoint', None, 'Location of the checkpoint to evaluate.')
```
mingxingtan, line 36: You can use yapf to format your code (max length is 80 characters)

LucasSloan: I tried running yapf on it and it went to 4 space indentation - is that correct?

efficientdet/keras/eval.py
LucasSloan: Done.

efficientdet/keras/eval.py
mingxingtan: No, it shouldn't. You need to use style=yapf, like here: https://github.com/google/automl/blob/master/efficientdet/g3doc/faq.md#21-how-can-i-format-my-code-for-prs

LucasSloan: Got it!

PR #590, opened 2020-07-12
initialize coco metric when filename is not present

etam103:
In evaluation function when filename is not present we need to initialize Coco there as well.
cc @mingxingtan


etam103:
> @googlebot I signed it!

PR #580, opened 2020-07-10
Modified EfficientDet

shivam-raj:
Heavier model based on efficientdet using a new FPN called QuadFPN (Basically a BiFPN and inverted BiFPN connected in parallel and a final FPN fusing features from parallel paths. For the architecture and checkpoints kindly refer to README of this branch. 

I have been experimenting on parallel fusion techniques and thus extracting more from backbone that one FPN can. 
A fully parallel FPN demo ,architecture and checkpoints are available on the master branch of my fork of this repo.
(Tho accuracy remains almost same but it is very inefficient and doesn't scale well. QuadFPN proves to be a better choice).

Another variant that even performed better was using unweighted QuadFPN (Unweighted parallel FPN connections and weighted connection for contribution of parallel  FPNs).Set fpn_quad_method='fastattn' and fpn_weight_method='sum' in hparams_config.py to use this. This provides more stable learning and better accuracy. Unfortunately my TFRC quota finished before I could do any further training.

PR #601, opened 2020-07-15
Add QuFPN per #580.

mingxingtan:
Also fix learning rate, add epoch summary, and add other code clean up

### DIFF
--- a/efficientdet/keras/fpn_configs.py
+++ b/efficientdet/keras/fpn_configs.py
@@ -1,0 +1,177 @@
+# Lint as: python3
+# Copyright 2020 Google Research. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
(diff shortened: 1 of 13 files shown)


mingxingtan:
@shivam-raj  With this PR, I am able to train D0 with fpn_name=qufpn, and improve validation accuracy from 34.0 to 35.0.  Could you please help review this PR? Thanks!

efficientdet/keras/fpn_configs.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def qufpn_config(min_level, max_level, weight_method=None, quad_method=None):
+  """A dynamic quad fpn config that can adapt to different min/max levels."""
+  # It extends the idea of BiFPN, and has four paths:
+  #   (up_down -> bottom_up) + (bottom_up -> up_down).
+  # See test for an example for level 2 and 7.
+  p = hparams_config.Config()
+  p.weight_method = weight_method or 'fastattn'
+  p.quad_method = quad_method or p.weight_method
```
shivam-raj, line 83: @mingxingtan 
p.quad_method shouldn't be ever set to unweighted sum (if p.weight_method is set to 'sum') , it produces bad results owing to equal contribution from the two parallel paths.
Actually you were right earlier, we don't really require two weight methods. p.quad_method is always set to 'fastattn' regardless of p.weight_method

```suggestion
  p.quad_method = 'fastattn'
```

shivam-raj:
@mingxingtan That's an even more impressive result !! 
Did you try training the bigger models while increasining the fpn_cell_repeats by 2 each time we scale the backbone?

mingxingtan:
> @mingxingtan That's an even more impressive result !!
> Did you try training the bigger models while increasining the fpn_cell_repeats by 2 each time we scale the backbone?

@shivam-raj  I will try to train a big model later when I get the resources.
I am submitingt this PR now to unblock other PRs.

PR #602, opened 2020-07-15
Add test time augmentation

LucasSloan:
Implements [Weighted Boxes Fusion](https://arxiv.org/pdf/1910.13302.pdf) for test time augmentation, currently only with 2 inferences per image (regular, and left-right flip).  I based my code on the [author's implementation of WBF](https://github.com/ZFTurbo/Weighted-Boxes-Fusion/blob/master/ensemble_boxes/ensemble_boxes_wbf.py).

I'm opening this as a draft because my current code has terrible performance.  I can eval the COCO validation set with d0 in 5-10 minutes normally, but with this code it take much longer (I don't know how long yet because the job hasn't actually finished, but probably 1 day+).  I am using a O(n^2) algorithm to cluster the boxes, but there are only 200 boxes per image, and we do something similar for the NMS with many more boxes, so I don't think that's it.  Any thoughts would be appreciated.

### DIFF
--- a/efficientdet/keras/eval.py
+++ b/efficientdet/keras/eval.py
@@ -39,6 +39,8 @@ flags.DEFINE_string('model_name', 'efficientdet-d0', 'Model name to use.')
 flags.DEFINE_string('model_dir', None, 'Location of the checkpoint to run.')
 flags.DEFINE_integer('batch_size', 8, 'Batch size.')
 flags.DEFINE_string('hparams', '', 'Comma separated k=v pairs or a yaml file')
+flags.DEFINE_boolean('enable_tta', False,
+                     'Use test time augmentation (slower, but more accurate).')
 FLAGS = flags.FLAGS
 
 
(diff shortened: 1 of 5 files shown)


LucasSloan:
Finally got a eval finished, with the left right flip I got an AP of 0.341.

mingxingtan:
> Finally got a eval finished, with the left right flip I got an AP of 0.341.

Great, thank you, Lucas!  What was the previous issue? (running very slow)

LucasSloan:
I haven’t fixed that yet, I just let it run for 16 hours.  I’ll be working
on the performance today.

mingxingtan:
FYI, I realized that using tf.function can speedup quite a lot. Here is an example: https://github.com/google/automl/blob/a488cf9d719d869f1e312e0b2e8cde243ab57103/efficientdet/keras/infer.py#L68

LucasSloan:
I tried using @tf.function, but it failed.  I need to rewrite the code in tensorflow primitives to let that happen.

LucasSloan:
Managed to improve performance by ~20x, now it takes under an hour to eval on the full validation set.

LucasSloan:
For some reason the algorithm changes to fix the speed issues reduced the performance as well - the current version of the code gets an AP of 0.339.

efficientdet/keras/eval.py
```diff
@@ -56,19 +59,32 @@ def main(_):
...
+    cls_outputs, box_outputs = model(images, training=False)
     detections = postprocess.generate_detections(config, cls_outputs,
                                                  box_outputs,
                                                  labels['image_scales'],
-                                                 labels['source_ids'])
-    evaluator.update_state(labels['groundtruth_data'].numpy(),
-                           detections.numpy())
+                                                 labels['source_ids'], False)
+
+    images_flipped = tf.image.flip_left_right(images)
```
mingxingtan, line 74: Could you make this as optional (maybe add a flag  like "enable_tta" or "enable_flip")?

efficientdet/keras/postprocess.py
```diff
@@ -354,20 +352,52 @@ def postprocess_per_class(params, cls_outputs, box_outputs, image_scales=None):
...
-  ]
+      params, cls_outputs, box_outputs, img_scales)
+
+  if flip:
+    _, width = utils.parse_image_size(params['image_size'])
+
+    img_ids_bs = tf.cast(tf.expand_dims(img_ids, -1), nms_scores_bs.dtype)
+    detections_bs = [
+        img_ids_bs * tf.ones_like(nms_scores_bs),
+        tf.expand_dims(img_scales, -1) * width - nms_boxes_bs[:, :, 3],
```
mingxingtan, line 371: I did not quite understand this logic. could you help clarify this? (maybe add some comments)?

efficientdet/keras/eval.py
LucasSloan: Done.

efficientdet/keras/postprocess.py
LucasSloan: Extracted a variable and added comments - better?

mingxingtan: Thanks for adding this comment. Looks good.

mingxingtan:
Thanks Lucas. Great to see simple flipping also helps. Would you also help add multi-scale testing?

LucasSloan:
Yep, that's my next project!

PR #605, opened 2020-07-17
fix infer bug

fsx950223:


efficientdet/keras/postprocess.py
```diff
@@ -214,10 +214,10 @@ def postprocess_combined(params, cls_outputs, box_outputs, image_scales=None):
           score_threshold=score_thresh,
           clip_boxes=False))
   nms_classes += CLASS_OFFSET
+  nms_boxes = clip_boxes(nms_boxes, params['image_size'])
```
mingxingtan, line 217: Why clip should happen before scaling the box? I thought we need to first scale and then clip?

fsx950223:
No, params['image_size'] is unscaled image size.

Mingxing <notifications@github.com> 于2020年7月18日周六 上午12:15写道：

mingxingtan:
You are right. Thanks for fixing this issue.

PR #637, opened 2020-07-30
fix display callback

### LINKED ISSUE #633: Keras test loss_vals have no total_loss
I think, the test_step function in train_lib is missing the total loss.

It should be added before the return.

```
  def test_step(self, data):
    (...)  
    loss_vals['loss'] = total_loss
    return loss_vals
```

fsx950223:
Fix #633


efficientdet/keras/train_lib.py
```diff
@@ -239,7 +247,8 @@ def get_callbacks(params, profile=False):
       params['model_dir'], verbose=1, save_weights_only=True)
   early_stopping = tf.keras.callbacks.EarlyStopping(
       monitor='val_loss', min_delta=0, patience=10, verbose=1)
-  return [tb_callback, ckpt_callback, early_stopping]
+  display_callback = DisplayCallback('./testdata/img1.jpg', os.path.join(params['model_dir'], 'train'))
```
mingxingtan, line 250: If user doesn't have ./testdata/img1.jpg,  would the training fail?
Maybe we should check if this file doesn't exist, then give a warning and continue training without display callback?

efficientdet/keras/train_lib.py
```diff
@@ -354,9 +365,17 @@ class EfficientDetNetTrain(efficientdet_keras.EfficientDetNet):
 
   see https://www.tensorflow.org/guide/keras/customizing_what_happens_in_fit
   """
+  def __init__(self, pattern=None, *args, **kwargs):
```
mingxingtan, line 368: Is there any issue for the old code that use config instead of a class member?

fsx950223: Thanks for pointing it out. I have removed it.

PR #643, opened 2020-08-01
Multi scale tta

LucasSloan:
Implements multi-scale tta + wbf's confidence scaling based on the number of boxes.

d0 + multiscale tta gets an AP of 37.1, an improvement of 1.3 over the highest single resolution used when paired with flip tta (35.8)

### DIFF
--- a/efficientdet/keras/eval.py
+++ b/efficientdet/keras/eval.py
@@ -21,6 +21,7 @@ import tensorflow as tf
 import coco_metric
 import dataloader
 import hparams_config
+import utils
 
 from keras import anchors
 from keras import efficientdet_keras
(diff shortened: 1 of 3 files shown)


efficientdet/keras/eval.py
```diff
@@ -46,51 +47,57 @@ def main(_):
...
   # Network
   model = efficientdet_keras.EfficientDetNet(config=config)
-  model.build((config.batch_size, None, None, 3))
+  model.build((config.batch_size, base_height, base_width, 3))
   model.load_weights(tf.train.latest_checkpoint(FLAGS.model_dir))
 
-  evaluator = coco_metric.EvaluationMetric(filename=config.val_json_file)
+  augmentations = [] 
+  if FLAGS.enable_tta:
+    for size_offset in (0, 128, 256):
```
mingxingtan, line 60: Maybe we should use some ratios like [0.5, 0.75, 1.0, 1.25, 1.5]?

efficientdet/keras/eval.py
```diff
@@ -46,51 +47,57 @@ def main(_):
...
+  base_height, base_width = utils.parse_image_size(config['image_size'])
 
   # Network
   model = efficientdet_keras.EfficientDetNet(config=config)
-  model.build((config.batch_size, None, None, 3))
+  model.build((config.batch_size, base_height, base_width, 3))
   model.load_weights(tf.train.latest_checkpoint(FLAGS.model_dir))
 
-  evaluator = coco_metric.EvaluationMetric(filename=config.val_json_file)
+  augmentations = []
```
mingxingtan, line 58: Add a comment:  # in format (height, width, flip)

mingxingtan:
Results look great. Thank you, Lucas!

efficientdet/keras/eval.py
LucasSloan: I'd like to finetune these more for the larger models, but for d0, it actively hurts to go below the default size (getting rid of that was the major fix I did).  Once I've got this code working internally so I can run the larger models, I'll experiment with these.

efficientdet/keras/eval.py
LucasSloan: Done.

PR #654, opened 2020-08-05
Fix preprocessing error in keras infer

wuhy08:
Somehow when a list of numpy arrays is passed to _preprocess, map_fn will use dim[0] of the first element in the list as n_static. A quick fix by converting the list to 4D tensor (for single image).


mingxingtan:
thanks for the fix!

PR #665, opened 2020-08-11
Speedup keras eval.

mingxingtan:


mingxingtan:
@LucasSloan This PR makes the keras eval much faster.
@fsx950223 Should we add eval to train_lib?

efficientdet/keras/eval.py
```diff
@@ -57,6 +57,13 @@ def main(_):
   model.build((config.batch_size, base_height, base_width, 3))
   model.load_weights(tf.train.latest_checkpoint(FLAGS.model_dir))
 
+  @tf.function
+  def f(imgs, labels, flip):
```
LucasSloan, line 61: Do we want to do this for infer.py as well?

Relatedly, the tta ensemble logic is still much slower than it should be - I tried to speed it up with tf.function, but it got slower due to a bunch of "retracing because shapes changed" stuff.  Do you know how to work around that?

mingxingtan: infer.py already has this logic: https://github.com/google/automl/blob/c35c3ec6ccd4e7778af413256060b49993770288/efficientdet/keras/infer.py#L64

PR #672, opened 2020-08-14
add ckpt_path to export best model

Samjith888:
Added the ckpt_path for exporting the best model while model conversion


mingxingtan:
thanks!

PR #677, opened 2020-08-14
Separating input and output model checkpoint, avoiding file overwriting.

wuhy08:
Currently, the flag `model_dir` is used for both loading and saving the checkpoint file in the keras training script. This will lead to a problem of overwriting the original checkpoint.

For example, if `model_dir=ckpt/efficientdet-d0`, the script will look for the latest checkpoint file inside the directory, where the filename is `model.*` if the pretrain weight downloaded from the repo is used. During training, checkpoint files are saved at `ckpt/efficientdet-d0.*`, at the same level of the original `model_dir`. After training is finished, the final checkpoint files called `model.*` are saved inside `model_dir`, potentially overwriting the original checkpoint, which might be intended to be kept.

With this PR, another flag called `ckpt` is created to take the directory of pretrained checkpoint directory and `model_dir` is continued to be used as output folder. 

Some other behavior changes including:

* Creating `model_dir` if not exist
* Intermediate checkpoint file is saved inside `model_dir`, named `ckpt.*`
* Final checkpoint file is also saved inside `model_dir`, named `ckpt-final.*`

Future work can include:
* Have the option of keeping multiple intermediate checkpoint files

### DIFF
--- a/efficientdet/keras/train.py
+++ b/efficientdet/keras/train.py
@@ -62,6 +62,9 @@ flags.DEFINE_bool(
     'and this flag has no effect.')
 flags.DEFINE_string('model_dir', None, 'Location of model_dir')
 
+flags.DEFINE_string('ckpt', None,
+                    'Start training from this EfficientDet checkpoint.')
+
 flags.DEFINE_string(
     'hparams', '', 'Comma separated k=v pairs of hyperparameters or a module'
     ' containing attributes to use as hyperparameters.')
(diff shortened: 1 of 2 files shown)


efficientdet/keras/train.py
```diff
@@ -62,6 +62,9 @@
     'and this flag has no effect.')
 flags.DEFINE_string('model_dir', None, 'Location of model_dir')
 
+flags.DEFINE_string('ckpt', None,
```
fsx950223, line 65: pretrained_ckpt

wuhy08: Did you mean changing `ckpt` to `pretrained_ckpt`?

fsx950223: yes

wuhy08: BTW, the reason I used `ckpt` is because the term was used in tf 1.x training.
See:
https://github.com/google/automl/blob/3958f1b68a02b5e67075e6c45ecc2376131f4165/efficientdet/main.py#L53-L54

wuhy08: Modified as suggested.

wuhy08:
Thank you, @fsx950223 !

PR #678, opened 2020-08-14
Fix an issue which prevents EarlyStopping callback

wuhy08:
When `val_loss` is used as metric for `EarlyStopping` monitoring, it only takes in scalar for comparison, raising the following error: `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`, since `loss_vals['loss']` is a 1st order tensor. A simple fix is to only add `det_loss[0]` to `total_loss`.

### DIFF
--- a/efficientdet/keras/train_lib.py
+++ b/efficientdet/keras/train_lib.py
@@ -508,7 +508,8 @@ class EfficientDetNetTrain(efficientdet_keras.EfficientDetNet):
       total_loss = reg_l2loss
       loss_vals = {}
       if 'object_detection' in self.config.heads:
-        det_loss = self._detection_loss(cls_outputs, box_outputs, labels, loss_vals)
+        det_loss = self._detection_loss(cls_outputs, box_outputs, labels,
+                                        loss_vals)
         total_loss += det_loss
       if 'segmentation' in self.config.heads:
         seg_loss_layer = self.loss['seg_loss']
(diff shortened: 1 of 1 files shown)


efficientdet/keras/train_lib.py
```diff
@@ -562,7 +562,7 @@ def test_step(self, data):
     if 'object_detection' in self.config.heads:
       det_loss = (
           self._detection_loss(cls_outputs, box_outputs, labels, loss_vals))
-      total_loss += det_loss
+      total_loss += det_loss[0]
```
fsx950223, line 565: It should be ```tf.math.reduce_sum(det_loss)``` and keep total loss same as train step.

wuhy08: However, in `_detection_loss`:
https://github.com/google/automl/blob/3958f1b68a02b5e67075e6c45ecc2376131f4165/efficientdet/keras/train_lib.py#L478-L480
which means the first term in the return value is already the weighted sum of the rest terms. Wouldn't applying `tf.math.reduce_sum` make the summation duplicate?

fsx950223: Yes, you're right. Change total_loss += det_loss[0] too.

fsx950223: I prefer to change _detection_loss function:
```return total_loss```

wuhy08: Modified as you suggested.

fsx950223:
You should follow our code style

wuhy08:
It is been yapf'ed. Thanks.

fsx950223:
LGTM, Thanks.

PR #693, opened 2020-08-18
:wrench: var_freeze pattern warning if none variables matched.

kartik4949:
Abseil logging Warning if no variables matched with  given var_freeze pattern.


mingxingtan:
Thanks for the fix.

PR #701, opened 2020-08-20
Added Mosaic Augmentation in core TF

kartik4949:
This is mosaic augmentation module in tf2,currently this augmentation doesnt preserve aspect ratio of original images passed
also it only supports 4 images to build mosaic.
Note: Tested in TF2 with eager execution and static graph.

### DIFF
--- a/efficientdet/aug/mosaic.py
+++ b/efficientdet/aug/mosaic.py
@@ -22,7 +22,7 @@ class Mosaic:
     """Mosaic.
         Mosaic Augmentation class.
         Notes:- 1. Mosaic sub images will not be preserving aspect ratio of original images passed.
-                2. Tested on eager tensor for now, future will make compatible with static graphs too.
+                2. Tested on static graphs and eager  execution.
                 3. This Implementation of mosaic augmentation is tested in tf2.x.
     """
     def __init__(
@@ -103,13 +103,12 @@ class Mosaic:
 
         """
         return [
-            box[0] * mosaic_image.shape[1] / image.shape[1],
-            box[1] * mosaic_image.shape[0] / image.shape[0],
-            box[2] * mosaic_image.shape[1] / image.shape[1],
-            box[-1] * mosaic_image.shape[0] / image.shape[0],
+            box[0] * tf.shape(mosaic_image)[1] / tf.shape(image)[1],
+            box[1] * tf.shape(mosaic_image)[0] / tf.shape(image)[0],
+            box[2] * tf.shape(mosaic_image)[1] / tf.shape(image)[1],
+            box[-1] * tf.shape(mosaic_image)[0] / tf.shape(image)[0],
         ]
 
-    @tf.function
     def _scale_images(self, images, mosaic_divide_points: Tuple) -> Tuple:
         """_mosaic.
             Scale Sub Images.
@@ -135,6 +134,7 @@ class Mosaic:
             mosaic_image_bottomright,
         )
 
+    @tf.function
     def _mosaic(self, images, boxes, mosaic_divide_points):
         """_mosaic.
             Builds mosaic of provided images.
@@ -178,8 +178,8 @@ class Mosaic:
         _num_boxes = boxes[1].shape[0]
         idx_tp = tf.constant([[1], [3]])
         update_tp = [
-            [mosaic_image_topleft.shape[0]] * _num_boxes,
-            [mosaic_image_topleft.shape[0]] * _num_boxes,
+            [tf.shape(mosaic_image_topleft)[0]] * _num_boxes,
+            [tf.shape(mosaic_image_topleft)[0]] * _num_boxes,
         ]
         mosaic_box_topright = tf.transpose(
             tf.tensor_scatter_nd_add(mosaic_box_topright, idx_tp, update_tp)
@@ -193,11 +193,12 @@ class Mosaic:
             ),
             boxes[2],
         )
+
         _num_boxes = boxes[2].shape[0]
         idx_bl = tf.constant([[0], [2]])
         update_bl = [
-            [mosaic_image_topleft.shape[1]] * _num_boxes,
-            [mosaic_image_topleft.shape[1]] * _num_boxes,
+            [tf.shape(mosaic_image_topleft)[1]] * _num_boxes,
+            [tf.shape(mosaic_image_topleft)[1]] * _num_boxes,
         ]
         mosaic_box_bottomleft = tf.transpose(
             tf.tensor_scatter_nd_add(mosaic_box_bottomleft, idx_bl, update_bl)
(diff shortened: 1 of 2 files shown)


kartik4949:
@mingxingtan can you review the format and what extra tests needed to add.

efficientdet/aug/mosaic_test.py
```diff
@@ -0,0 +1,46 @@
...
+
+class MosaicTest(tf.test.TestCase):
+
+    def __init__(self,*args,**kwargs):
+        super().__init__(*args,**kwargs)
+        self.output_size = (512,512)
+        self.mosaic = Mosaic(out_size=self.output_size)
+        tf.random.set_random_seed(111111)
+
+    def test_mosaic_boxes(self):
```
LucasSloan: Can you add a test that uses 4 tiny images?  Ie 4x4 images of all zeroes, ones, twos, etc.

efficientdet/aug/mosaic.py
```diff
@@ -0,0 +1,261 @@
...
+            Scale Sub Images.
+
+        Args:
+            images: original single images to make mosaic.
+            mosaic_divide_points: Points to build mosaic around on given output size.
+        Returns:
+            (tuple)
+            Scaled Mosaic sub images.
+        """
+        x, y = mosaic_divide_points[0][0], mosaic_divide_points[1][0]
```
LucasSloan: The mosaic algorithm doesn't do this resizing - it does cropping.  See the darknet implementation [here](https://github.com/AlexeyAB/darknet/blob/master/src/data.c#L1907).

kartik4949: @LucasSloan we can make mosaic of original image and then crop the mosaic with output_size

efficientdet/aug/mosaic_test.py
kartik4949: @LucasSloan Sure i can

efficientdet/aug/mosaic.py
kartik4949: @LucasSloan FYI ,this is initial version of mosaic we can test with resizing ,if it doesnt help ,I can surely make the cropping version.

efficientdet/aug/mosaic_test.py
LucasSloan: Also add a test for the bounding box munging.

efficientdet/aug/mosaic.py
LucasSloan: Resizing is already a knob that we have in the augmentation toolkit here.  If you do mosiac via resizing, you're reducing the effective size of everything by a factor of 2.  It's not good to have this interaction between different augmentations - please redo this as a crop.

kartik4949: you can control the mosaic image shape via output_size

kartik4949: so lets say ,u have 4 images with shape (512x512) you can make a mosaic of them with output size (512,512) ,right?

efficientdet/aug/mosaic_test.py
kartik4949: @LucasSloan can you add munging test.

kartik4949:
@mingxingtan can we add this for now as a template for mosaic augmentation, we can progress on it.

efficientdet/aug/mosaic_test.py
kartik4949: added tiny image test.

kartik4949:
@LucasSloan @mingxingtan  i can make mosaic of  images without resizing them and later on will use the resize_image_crop and resize_boxes_and_crop function which will make random crops on the mosaic image ,
this is solution will work and i will add this in my next PR ,which  will make it exactly like darknet .

itsliupeng:
I wonder how to get the left 3 random  images from the tfrecord file.

efficientdet/aug/mosaic.py
```diff
@@ -0,0 +1,262 @@
...
+      out_size: Tuple = (680, 680),
+      n_images: int = 4,  #Currently four images are supported!
+      _minimum_mosaic_image_dim: int = 25,
+  ):
+    """__init__.
+
+        Args:
+            out_size: output mosaic image size.
+            n_images: number images to make mosaic
+            _minimum_mosaic_image_dim: minimum percentage of out_size dimension should the mosaic be. i.e if out_size is (680,680) and _minimum_mosaic_image_dim is 25 , minimum mosaic sub images dimension will be 25 % of 680
```
mingxingtan, line 41: This line seems too long.

Could you use yafs to format your code: https://github.com/google/automl/blob/master/efficientdet/g3doc/faq.md#21-how-can-i-format-my-code-for-prs

mingxingtan:
@fsx950223 has done some augmentation extension before, you might also want to take a look.

efficientdet/aug/mosaic.py
kartik4949: yapf --in-place --style='{based_on_style=yapf}' aug/mosaic.py

kartik4949: i did use this

kartik4949:
> I wonder how to get the left 3 random images from the tfrecord file.
> In yolov5, it samples from all the train data https://github.com/ultralytics/yolov5/blob/master/utils/datasets.py#L614

@itsliupeng Good point, I need to sample 4 images from batches.

fsx950223:
> I wonder how to get the left 3 random images from the tfrecord file.
> In yolov5, it samples from all the train data https://github.com/ultralytics/yolov5/blob/master/utils/datasets.py#L614

batch(3).map(process)

fsx950223:
LGTM

efficientdet/aug/mosaic.py
kartik4949: can you format from your end? we can squash  the commits later.

mingxingtan: Oh, I think because it is in a comment start with """, then the formater couldn't reformat it.  Could you just manually make them fit "80 characters per line" rule? (should be fairly simple.

kartik4949: sure.

kartik4949: @mingxingtan Resolved!. Thanks for pointing out, we can merge it now.

innat:
@kartik4949 cc. @mingxingtan @fsx950223 I was curious to apply mosaic augmentation not for detection but for the classification tasks. Could you please give a look at [it](https://stackoverflow.com/questions/65181294/how-to-create-class-label-for-mosaic-augmentation-in-image-classification)?

kartik4949:
> @kartik4949 cc. @mingxingtan @fsx950223 I was curious to apply mosaic augmentation not for detection but for the classification tasks. Could you please give a look at [it](https://stackoverflow.com/questions/65181294/how-to-create-class-label-for-mosaic-augmentation-in-image-classification)? 

Hi, 
Mosaic on classification would be asking to classify dog cat problem when both dog and cat are present in the image 

So in this case the label for mosaic augmented image for classification 
"Can be" as followed 

 1. You can modify the CutMix algorithm slightly based on no of classes 
2. You can also go with dominant class strategy (Im skeptical with this though)

Also mosaic is in a way a more aggressive form of cutmix without overlap in case of classification.

PR #706, opened 2020-08-26
add option run_epoch_in_child_process

### LINKED ISSUE #224: CPU memory in use increases during Training
I followed your instruction "6. Train on PASCAL VOC 2012 with backbone ImageNet ckpt" and "7. Finetune on PASCAL VOC 2012 with detector COCO ckpt".  
In both cases, I encountered a problem that CPU memory in use increases during training.  
The following is the result of free --giga -s 1800, during the training: 
``` 
             total        used        free      shared  buff/cache   available
Mem:             65           1          19           0          44          63
Swap:             2           0           2

              total        used        free      shared  buff/cache   available
Mem:             65           6          13           0          44          58
Swap:             2           0           2

              total        used        free      shared  buff/cache   available
Mem:             65           8          11           0          44          55
Swap:             2           0           2

              total        used        free      shared  buff/cache   available
Mem:             65          10           9           0          45          54
Swap:             2           0           2
...
...
```
And the process was eventually killed when the CPU memory in use reached its maximum 
(after about 18 hours for D0;  the time to be killed is shorter when I used D2, about 12 hours).
My OS is Ubuntu 18.04.

I couldn't find the reason by myself.  I am afraid that this problem might affect the overall training time, and would appreciate if you could modify this matter in future revision.  Thanks.

### LINKED ISSUE #688: Memory leak in train_and_eval mode
I've been observing that memory (RAM) usage is creeping up in train_and_eval mode. I had to stop training and launch it again every ~12 hours to prevent the process being killed by the system due to memory overuse. I have found based on this [thread](https://github.com/tensorflow/tensorflow/issues/32052) that this is a general problem with Estimator.train inside a loop through epochs. There are suggestions that google TCMalloc library helps. TCMalloc seems to have alleviated the problem a little bit but did not solve it for me. After 12 hours running before TCMalloc I had ~70% eaten, after TCMalloc ~50% eaten.

Below is an analysis comparing the memory heap in the middle of execution and in the end of execution and showing only the difference (run by command: google-pprof --text --base=/tmp/profile.0500.heap main.py /tmp/profile.0676.heap | head -n 15)

Total: 860.1 MB
   193.4  22.5%  22.5%    216.2  25.1% google::protobuf::DynamicMessage::New
   170.0  19.8%  42.3%    170.0  19.8% std::vector::emplace_back
   102.5  11.9%  54.2%    102.5  11.9% std::string::_Rep::_S_create
    81.7   9.5%  63.7%     81.7   9.5% 00000000005b7fa4
    64.7   7.5%  71.2%    121.5  14.1% tensorflow::ShapeRefiner::AddNode
    46.7   5.4%  76.6%     51.6   6.0% tensorflow::shape_inference::InferenceContext::InferenceContext
    33.4   3.9%  80.5%     33.4   3.9% google::protobuf::internal::RepeatedPtrFieldBase::InternalExtend@d4de0
    29.7   3.5%  84.0%     29.7   3.5% google::protobuf::internal::InternalMetadata::mutable_unknown_fields_slow
    21.4   2.5%  86.4%     20.5   2.4% google::protobuf::Arena::CreateMaybeMessage
    9.5   1.1%  94.1%     25.9   3.0% tensorflow::NodeDef::NodeDef
    20.7   2.4%  88.9%     41.5   4.8% tensorflow::Graph::AddNode
    14.3   1.7%  90.5%     14.3   1.7% google::protobuf::Map::InnerMap::CreateEmptyTable
    11.1   1.3%  91.8%     12.5   1.5% google::protobuf::Map::operator[]
    10.1   1.2%  93.0%     22.8   2.6% google::protobuf::internal::DynamicMapField::DynamicMapField
    9.5   1.1%  94.1%     25.9   3.0% tensorflow::NodeDef::NodeDef

NikZak:
Closes #688, closes #224

### DIFF
--- a/efficientdet/hparams_config.py
+++ b/efficientdet/hparams_config.py
@@ -276,7 +276,7 @@ def default_detection_configs():
   # A temporary flag to switch between legacy and keras models.
   h.use_keras_model = True
 
-  # in case of RAM memory leak bug try with True
+  # in case of CPU memory leak bug try with True
   h.each_epoch_in_separate_process = False
 
   # unused.
--- a/efficientdet/hparams_config.py
+++ b/efficientdet/hparams_config.py
@@ -276,8 +276,8 @@ def default_detection_configs():
   # A temporary flag to switch between legacy and keras models.
   h.use_keras_model = True
 
-  # in case of CPU memory leak bug try with True
-  h.each_epoch_in_separate_process = False
+  # rectifies CPU memory leak
+  h.each_epoch_in_separate_process = True
 
   # unused.
   h.resnet_depth = 50
--- a/efficientdet/hparams_config.py
+++ b/efficientdet/hparams_config.py
@@ -276,9 +276,6 @@ def default_detection_configs():
   # A temporary flag to switch between legacy and keras models.
   h.use_keras_model = True
 
-  # rectifies CPU memory leak
-  h.run_epoch_in_child_process = True
-
   # unused.
   h.resnet_depth = 50
   return h
--- a/efficientdet/hparams_config.py
+++ b/efficientdet/hparams_config.py
@@ -276,6 +276,9 @@ def default_detection_configs():
   # A temporary flag to switch between legacy and keras models.
   h.use_keras_model = True
 
+  # in case of RAM memory leak bug try with True
+  h.each_epoch_in_separate_process = False
+
   # unused.
   h.resnet_depth = 50
   return h
--- a/efficientdet/hparams_config.py
+++ b/efficientdet/hparams_config.py
@@ -277,7 +277,7 @@ def default_detection_configs():
   h.use_keras_model = True
 
   # rectifies CPU memory leak
-  h.each_epoch_in_separate_process = True
+  h.run_epoch_in_child_process = True
 
   # unused.
   h.resnet_depth = 50
--- a/efficientdet/main.py
+++ b/efficientdet/main.py
@@ -105,9 +105,9 @@ flags.DEFINE_integer(
 # for train_and_eval mode
 flags.DEFINE_bool(
     'each_epoch_in_separate_process', False,
-    'this helps with RAM memory leak as inside the epoch loop it runs every'
-    'train_and_eval function in a separate process and clears memory after'
-    'the process end.\n'
+    'This option helps to rectify RAM memory leak. If set to True then every '
+    'epoch iteration is run in a separate process '
+    'for train_and_eval mode and the memory is cleared after each epoch.\n'
     'Drawback: you need to kill 2 processes instead of one if '
     'you want to interrupt training')
 
--- a/efficientdet/main.py
+++ b/efficientdet/main.py
@@ -14,6 +14,7 @@
 # ==============================================================================
 """The main training script."""
 import os
+import multiprocessing
 from absl import app
 from absl import flags
 from absl import logging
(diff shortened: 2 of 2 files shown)


NikZak:
@googlebot I signed it!

efficientdet/main.py
```diff
@@ -347,6 +357,13 @@ def _eval(steps):
       eval_results = _eval(eval_steps)
       ckpt = tf.train.latest_checkpoint(FLAGS.model_dir)
       utils.archive_ckpt(eval_results, eval_results['AP'], ckpt)
+    for e in range(current_epoch + 1, config.num_epochs + 1, epochs_per_cycle):
+      if FLAGS.each_epoch_in_separate_process:
+        p = multiprocessing.Process(target=lambda: run_train_and_eval(e))
```
mingxingtan, line 362: Thank you! This is a brilliant idea.  Does it solve the memory leak issue?

NikZak: Yes, it does

efficientdet/hparams_config.py
```diff
@@ -276,6 +276,9 @@ def default_detection_configs():
   # A temporary flag to switch between legacy and keras models.
   h.use_keras_model = True
 
+  # in case of CPU memory leak bug try with True
+  h.each_epoch_in_separate_process = False
```
mingxingtan, line 280: Maybe make this True in default? Any drawback for having separate process?

NikZak: I don't see any major drawbacks. I agree that True in default is a good idea. Things to consider 1) it has not been tested on Windows 2) it has not been tested in cloud 3) it has not been tested with TPU. Having said that, I think it should work in all environments and there is little risk in these issues

mingxingtan: Many people don't know this option and still have OOM issues if it is False.
It seems safe to me, but if it causes any unexpected issues later, we can revert it back to False.

NikZak: commited h.each_epoch_in_separate_process = True in default

kartik4949:
@mingxingtan @NikZak if we have 300 epochs how many  processes will be made?

NikZak:
One process is made per one epoch. Every time an epoch ends the process also ends. There are always only two simultaneously running processes - one for the main.py and one created inside the epoch loop

kartik4949:
> One process is made per one epoch. Every time an epoch ends the process also ends. There are always only two simultaneously running processes - one for the main.py and one created inside the epoch loop

got it!

kartik4949:
@NikZak every time epoch  ends ,how the process ends?

efficientdet/main.py
kartik4949: i think you need to terminate process explicitly.

NikZak:
Once the epoch is finished the process is terminated with exitcode 0

efficientdet/main.py
NikZak: Why do you want to terminate rather than wait for the process to finish? I can add p.close() before p.join(). I tested it and it is working. But I'm not sure what is the use for p.close() in here. Is it something you are looking for?

kartik4949: nah we cool , you already explained it below.

efficientdet/hparams_config.py
```diff
@@ -276,6 +276,9 @@ def default_detection_configs():
   # A temporary flag to switch between legacy and keras models.
   h.use_keras_model = True
 
+  # rectifies CPU memory leak
+  h.run_epoch_in_child_process = True
```
mingxingtan, line 280: I think you don't need this config, since you already have a flag in main.py?

mingxingtan: config is used to pass around different files, but since only main.py needs this, making it a FLAG in main.py is better.

NikZak: Thanks. I was also wondering why need to set the default in two places. Corrected now

Issue #705, opened 2020-08-25
Saving keras model returns InaccessibleTensorError


robert-hardwick:
I have added

model.save('/path/to/saved_model') 

to keras/infer.py 

I am able to load weights into model and perform predictions, but i cannot save the model as per the following error.

tensorflow.python.framework.errors_impl.InaccessibleTensorError: The tensor 'Tensor("blocks_15/Identity_2:0", shape=(None, 16, 16, 320), dtype=float32)' cannot be accessed here: it is defined in another function or code block. Use return values, explicit Python locals or TensorFlow collections to access it. Defined in: FuncGraph(name=efficientnet-b0_layer_call_and_return_conditional_losses, id=139902875332736); accessed from: FuncGraph(name=_layer_call_and_return_conditional_losses, id=139902877748360).

Anyone have any tips / suggestions to debug?

robert-hardwick:
Looks like the issue is related to https://github.com/tensorflow/tensorflow/issues/37444

and the problem code is here

https://github.com/google/automl/blob/master/efficientdet/keras/efficientdet_keras.py#L820

kartik4949:
@fsx950223

robert-hardwick:
Just trying to export a model that I have trained for segmentation and object detection to a saved_model format, but i have this error if i use keras scripts directly. 

If i used model_main.py --run_mode saved_model i receive other errors:

```
2020-08-26 16:10:25.126783: W tensorflow/core/framework/op_kernel.cc:1774] OP_REQUIRES failed at save_restore_v2_ops.cc:205 : Not found: Key box_net/box-0-bn-3/beta not found in checkpoint
WARNING:tensorflow:Restoring an object-based checkpoint using a name-based saver. This may be somewhat fragile, and will re-build the Saver. Instead, consider loading object-based checkpoints using tf.train.Checkpoint().
W0826 16:10:25.142078 139644081489728 saver.py:1320] Restoring an object-based checkpoint using a name-based saver. This may be somewhat fragile, and will re-build the Saver. Instead, consider loading object-based checkpoints using tf.train.Checkpoint().
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/client/session.py", line 1365, in _do_call
    return fn(*args)
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/client/session.py", line 1350, in _run_fn
    target_list, run_metadata)
[...]

PR #708, opened 2020-08-27
fix keras export saved model

fsx950223:
Fix https://github.com/google/automl/issues/705


efficientdet/keras/infer.py
```diff
@@ -60,12 +63,25 @@ def main(_):
...
+      super().__init__()
+      self.model = model
+
+    @tf.function
+    def f(self, imgs):
+      return self.model(imgs, training=False, post_mode='global')
 
   imgs = tf.convert_to_tensor(imgs, dtype=tf.uint8)
-  boxes, scores, classes, valid_len = f(imgs)
+  export_model = ExportModel(model)
```
mingxingtan, line 76: It is okay for this CL, but maybe later we should provide all TF2 savedmodel as well.

fsx950223: How about export saved model to TensorFlow hub

mingxingtan:
Hi @fsx950223 This PR breaks the following tests:

python -m backbone.efficientnet_builder_test
python -m backbone.efficientnet_model_test
python -m efficientdet_arch_test

Could you help fix those tests? Thanks!

PR #711, opened 2020-08-28
Gradient checkpointing

NikZak:
@mingxingtan 
@kartik4949 

Closes #85 , closes #368, closes #459, closes #737
Depends on #716

This is a slightly augmented version of [gradient check-pointing algorithm](https://github.com/cybertronai/gradient-checkpointing) for efficientdet network. It also includes porting of [graph editor](https://github.com/tensorflow/tensorflow/tree/r1.15/tensorflow/contrib/graph_editor) from tensorflow.contrib 1.15

This is an experimental option. It helps to save GPU memory while training. 

As input, you need to provide a list of strings that would indicate which layers of the network to use to save checkpoints.

When this option is used the standard tensorflow.python.ops.gradients method is being replaced with a custom method. The parameters that you use are important and this requires further optimization. It takes time to reassemble the computation graph with new checkpoints and this operation is not multi-threaded at the moment - this could be improved. The graph reassembling only happens once per epoch in the beginning of the training epoch. Another improvement that could be made is caching the graph between epochs (which may not be straightforward given that every epoch runs in a separate process)

You have to provide a list of strings. These strings will be searched in the tensors name and only those tensors that match will be kept as checkpoints.

['L2Loss', 'entropy', 'FusedBatchNorm', 'Switch', 'dropout', 'Cast'] layers are always removed.

**gradient_checkpointing: ["Add"]** (leave only tensors with Add in the name) is an option that has been tested  and works reasonably well: 

1.     For d4 network with batch-size of 1 (mixed precision enabled) it takes only 1/3.2 of memory with roughly 32% slower computation with this option enabled.
2.     It also allows to train a d6 network with batch size of 2 (mixed precision enabled) on a 11Gb GPU which is impossible without this option

There were also some logging improvements added and in particular memory logging for Nvidia GPU (disabled by default) (for a single GPU at the moment as I don't have a multi GPU machine to test multiple GPUs)

I suggest to add this option as it does not break the main process flow. The memory improvement for GPU is very substantial but could be further optimized and improved through providing right params or changing the algorithm

PR #716, opened 2020-08-31
Gradient checkpointing libraries

NikZak:
These are the libraries for #711

This PR does not change the existing files

It contains graph_editor and memory_saving_gradients.py used in #711


efficientdet/graph_editor/BUILD
```diff
@@ -0,0 +1,162 @@
+# Description:
+#   contains parts of TensorFlow that are experimental or unstable and which are not supported.
+
```
fsx950223, line 3: Maybe we could remove BUILD files?

NikZak: Sure, removed

efficientdet/gradient_checkpointing/LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+The MIT License
+
+Copyright (c) 2018 OpenAI (http://openai.com)
```
mingxingtan, line 3: Hmm, we are not able to import code with such copyright.

kartik4949: @mingxingtan memory_saving_gradients is open source codebase (openai) ,cant we keep the copyright?

NikZak: I see. That's too bad. Is there any workaround? In the meantime I may try to ask the authors if there's a way to remove the copyright keeping mentioning their names. How can l I explain why it has to be removed?

NikZak: @mingxingtan
What if keep all the libraries as a separate repository? Then you don't need to copy them here, you would be able to use them with "pip install" and import. Is it a valid workaround?

NikZak: Great, thanks. Waiting for your resolution then

mingxingtan:
We have decided to merge this third_party code!
@NikZak Could you run yafq and pylint to reformat your code? Thanks!

fsx950223:
How about move it to ```third_party``` directory?

NikZak:
Thanks. I will yapf, pylint and adopt to latest changes

mingxingtan:
Good idea @fsx950223 !

@NikZak Could you create a third_party folder and move this code to third_party? You can slightly simply folder names, such as:

gradient_checkpointing  -> third_party/grad_checkpoint
graph_editor -> third_party/graph_edit

NikZak:
Ok, this is done. Please merge this one

efficientdet/third_party/graph_edit/edit.py
```diff
@@ -0,0 +1,223 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from third_party.graph_edit import reroute
+from third_party.graph_edit import select
+from third_party.graph_edit import subgraph
+from third_party.graph_edit import util
+from tensorflow.python.ops import array_ops as tf_array_ops
```
mingxingtan, line 25: It would be better to avoid importing tf internal stuffs?

For example, here you can simply import tensorflow as tf, and replace the tf_array_ops.placeholder as "tf.placeholder"?

efficientdet/third_party/graph_edit/examples/edit_graph_example.py
```diff
@@ -0,0 +1,53 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+import tensorflow as tf
+import third_party.graph_edit as ge
+
+FLAGS = tf.flags.FLAGS
```
mingxingtan, line 25: Looks like this FLAGS is not used anywhere. How about deleting this line?

efficientdet/third_party/graph_edit/tests/edit_test.py
```diff
@@ -0,0 +1,82 @@
...
+  """
+
+  def setUp(self):
+    """Set up."""
+    self.graph = ops.Graph()
+    with self.graph.as_default():
+      self.a = constant_op.constant([1., 1.], shape=[2], name="a")
+      with ops.name_scope("foo"):
+        self.b = constant_op.constant([2., 2.], shape=[2], name="b")
+        self.c = math_ops.add(self.a, self.b, name="c")
```
mingxingtan, line 43: If you simply import tensorflow, and use "tf.add" here; then it is much easier to read and maintain?

mingxingtan (commented):
Looks looks good to me.

mingxingtan (commented):
My biggest advice is to avoid importing tensorflow internal modules, which would make it much easier to read and maintain.

(internally, importing tensorflow internal modules is prohibited except very special cases.)

mingxingtan:
I realized the original graph_edit is package of tf.contrib, which is an internal tf library, so it might be fine in this case.

I tried to run the tests under third_party/graph_edit, but they all failed. Could you help fixed those tests? Thanks!

efficientdet/third_party/graph_edit/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""TensorFlow Graph Editor."""
+
+from __future__ import absolute_import
```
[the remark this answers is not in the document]
NikZak: Thanks. Removed

efficientdet/third_party/graph_edit/examples/edit_graph_example.py
NikZak: Thanks. Removed

NikZak:
The tests are running now. There were two special cases, testing tf.cond and tf.while in a copy of the graph, that did not pass. This is not needed for the purpose of efficientdet.

Please run the tests as described in the FAQ to avoid import errors
https://github.com/google/automl/blob/eb74c6739382e9444817d2ad97c4582dbe9a9020/efficientdet/g3doc/faq.md#22-how-can-i-run-all-tests

if you want to run with pytest make sure efficientdet folder is in PYTHONPATH

mingxingtan:
@NikZak 
Is this PR stable now? If so, I will try to import this PR, but you need to freeze this PR (don't submit new commits to this branch).

NikZak:
It's stable. OK, got it

efficientdet/third_party/grad_checkpoint/memory_saving_gradients.py
```diff
@@ -0,0 +1,519 @@
+"""Recompute tensorflow graph reducing GPU memory usage."""
+import logging
+import contextlib
+import time
+import sys
+from toposort import toposort
+import numpy as np
```
mingxingtan, line 7: Is it possible to replace this toposort with a function?  (We don't have this library internally).

NikZak: Ok. Done

NikZak: Ok. Done

LucasSloan:
Why are we vendoring these dependencies?  It looks like the answer is that there were a bunch of tweaks to the underlying libraries?

NikZak:
@LucasSloan
None of these libraries work with tf 2.x originally

Main changes:
Graph editing - minimal changes in the syntaxis, refactoring and some tweaks in the tests. 
Memory gradients - the algorithm of picking the gradients is substantially changed, the algorithm of gluing graph back together is left as is. Refactored. 
Nvgpu - a completely independent algorithm with same name as a pypi repo
Toposort - no changes, just refactoring

LucasSloan (approved):

fsx950223:
What's difference between ```debug_print``` and ```logging.debug```.

efficientdet/third_party/grad_checkpoint/memory_saving_gradients.py
```diff
@@ -0,0 +1,521 @@
...
+      logging.debug("ts_all: %s", len(ts_all))
+
+      # filter out all tensors that are inputs of the backward graph
+      with util.capture_ops() as bwd_ops:
+        tf_gradients(ys, xs, grad_ys, **kwargs)
+
+      bwd_inputs = [t for op in bwd_ops for t in op.inputs]
+      # list of tensors in forward graph that is in input to bwd graph
+      ts_filtered = list(set(bwd_inputs).intersection(ts_all))
+      debug_print("Using tensors %s", ts_filtered)
```
fsx950223, line 172: What's difference between debug_print and logging.debug.
Maybe we should use absl.logging.
Could you add some test cases for memory_saving_gradients? Thanks.

NikZak: Hi @fsx950223  
1) Regarding debug_print you can refer to the function description 
```python 
  """Like logger.log, but also replaces all TensorFlow ops/tensors with their
  names. Sensitive to value of DEBUG_LOGGING, see enable_debug/disable_debug
  Usage:
    debug_print("see tensors %s for %s", tensorlist, [1,2,3])
  """
```
If you want to log the tensors themselves debug_print is a good option (but it may be quite wordy). If no tensors standard logging.debug would do
2) Thanks for the suggestion. What does this library do? I do not want to add a dependency on any other third party library as it may lead to necessity to vendor them to this repo due to google internal rules. Then need refactoring and so on. Could be a neverending cycle 
3) Regarding testing. More testing coverage is always good. Probably I will add some tests at a later stage (maybe it should be part of a bigger integration test for this repo as gradient checkpointing is a major switch) and probably I will add some optimizations to the algorithm. However, this pull request is stable and it is on the way to be merged so unless it is a bug I don't want to make any changes

fsx950223: We always use absl.logging in the repo.

NikZak: Thanks, as it is not a bug, I will rectify after merging this pull request

mingxingtan:
@NikZak @fsx950223 @LucasSloan 
I am merging this PR now, followed by another PR to fix some format issues.

Issue #676, opened 2020-08-14
No detections after training on custom dataset


nemtiax:
I'm interested in training from scratch (or rather from an efficientNet backbone) rather than fine-tuning.  After training on my dataset, the training loss values looked good, but evaluation showed no detections, and AP=0.  Even running the trained model on images from the training set gave no detections.  I've found a few other issues (#521, #499 and #68) reporting similar problems, but following the suggestions in those issues hasn't worked so far.

To try to get a better handle on what's going wrong, I'm now trying to learn a very small dataset consisting of only eight COCO images using only the person class.  My hope is that it should be easy for the model to quickly overfit this tiny dataset, and I'll be able to quickly see it learn to make detections on the training images.  However, even on this reduced dataset, I'm still getting no detections.

My training command looks like this:

```
python main.py --mode=train --training_file_pattern=tfrecord/train*.tfrecord --validation_file_pattern=tfrecord/train*.tfrecord --model_name=efficientdet-d0 --model_dir=efficientdet-d0-single_batch  --backbone_ckpt=noisy_student_efficientnet-b0  --train_batch_size=8 --num_examples_per_epoch=256 --num_epochs=500 --hparams=person_only_config.yaml
```
Note that I'm using a efficientnet backbone from here: https://github.com/tensorflow/tpu/tree/master/models/official/efficientnet

My yaml config file specifies that I want only the person class:

```
num_classes: 2
label_map: {1: person}
```

I set num_classes to 2 following the advice in other issues.  I have also tried with num_classes=1, which gives the same results.

My training loss curves look like this:

<img width="1173" alt="Screen Shot 2020-08-14 at 11 16 05 AM" src="https://user-images.githubusercontent.com/6548233/90264776-972eca80-de1f-11ea-84c9-09b05375a46f.png">

This certainly looks reasonable at first glance, although I might have expected them to reach (near-)zero given that the dataset literally has only 8 images, and the model certainly has sufficient capacity to simply memorize it.  It seems at least plausible that the reduction in loss represents the initial model outputting nonsense boxes (and getting penalized severely), and then slowly learning to just output nothing (and so only getting small penalties for missing the few boxes that should be there).  

[filtered_train.txt](https://github.com/google/automl/files/5075376/filtered_train.txt)

Above is my reduced annotation set (note that it's been renamed from .json to .txt so github will accept it).  It's certainly possible I've made some mistake when generating it, but it appears to my eyes to have the right annotations.  The corresponding COCO train images are:

```
[...]

PR #720, opened 2020-09-01
add gradients test

fsx950223:
Related #676.
I got same gradients when I ran test cases. Maybe there is something wrong with efficientdet detection model.


fsx950223:
I found keras model got different update ops with legacy model before and I fix it in the PR.

fsx950223:
cc @mingxingtan

PR #722, opened 2020-09-01
Sync internal changes.

mingxingtan:


efficientdet/dataloader.py
```diff
@@ -394,7 +394,10 @@ def __call__(self, params):
 
     # Prefetch data from files.
     def _prefetch_dataset(filename):
-      dataset = tf.data.TFRecordDataset(filename).prefetch(1)
+      if params.get('dataset_type', None) == 'sstable':
```
kartik4949, line 397: 'sstable'?

efficientdet/dataloader.py
```diff
@@ -406,10 +409,16 @@ def _prefetch_dataset(filename):
       dataset = dataset.shuffle(64)
```
kartik4949, line 409: tf.dataset.list_files already shuffles the dataset

mingxingtan: That's to shuffle the files, this is to shuffle the examples?

efficientdet/dataloader.py
mingxingtan: Yeah, it is an internal experimental feature.

efficientdet/dataloader.py
kartik4949: oh it is only usefull if we have one single record file,else i dont see any use of it. because batch_size is one here.

kartik4949:
LGTM

fsx950223 (approved):

kartik4949 (approved):

efficientdet/det_model_fn.py
```diff
@@ -460,6 +475,8 @@ def metric_fn(**kwargs):
                   nms_configs['max_output_size'],
               ], tf.float32)
           detections_bs.append(detections)
+        detections_bs = postprocess.transform_detections(
```
LaurensHagendoorn, line 478: This detection_bs does not appear to be used elsewhere? so why bother calculating it? Also, shouldn't the detection_bs in line 497 also be transformed if this is used?

kartik4949: used in eval_metric

LaurensHagendoorn: where? pycharm says this variable is not used, I also don't see it passed anywhere?

kartik4949: line 503

LaurensHagendoorn: ohw, my bad, I edited out the coco stuff in my version. Thanks!

Issue #719, opened 2020-09-01
Memory Leak in "mode=train", strange evaluation curve


SiBensberg:
Thank you for your great work, at the moment I am experiencing 2 problems i want to solve.

1. The recent fix for the memory leak in "train_and_eval" worked well _(Thank you for that!)_. But I still have a memory leak in "train" mode. In about 4 hours the RAMs usage got up to 60gb and gets terminated by the system. I can restart the training without problems. So it is more an minor inconvenience. Am I doing something wrong? using `tf 2.2` + `Cuda 10.1.243`

2. The eval AP50 curve in tensorboard looks strange. The AP is about 0.022 which is not good. But the inference looks very good and way better than an AP50 of 0.022. Am I doing something wrong with the eval params?
![image](https://user-images.githubusercontent.com/55738346/91832295-85359000-ec45-11ea-844a-0f7c3c79e6b6.png)
![image](https://user-images.githubusercontent.com/55738346/91832597-e8bfbd80-ec45-11ea-8558-3d9db05d9b6d.png)

I made some minor changes to the training and loaded Efficientdet-D1 Coco weights.
I am using (**voc-config.yaml**)

```
num_classes: 5
var_freeze_expr: '(efficientnet)'
label_map: {1: blue, 2: yellow, 3: orange, 4: orange_big}
use_keras_model: False
anchor_scale: 1  #smallest Objects to detect are ~8px so an anchor scale of 1 should do it
num_scales: 5
aspect_ratios: [[1.0, 1.0], [0.634, 1.577], [1.577, 0.634]]
learning_rate: 0.015
lr_warmup_init: 0.0015
max_instances_per_image: 200     # got an error at one time
```
**eval:**
```
python3 main.py --mode=eval  \
    --model_name=efficientdet-d1  --model_dir=efficientdet-d1-sgd_anchor  \
    --validation_file_pattern=/home/val*.record  \ # Just one file
    --hparams=voc_config.yaml \
    --eval_batch_size=1 --eval_samples=5000
```

**train:**
```
python3 main.py --mode=train \
    --training_file_pattern=/home/train*.record \
    --validation_file_pattern=/home/val*.record \
    --model_name=efficientdet-d1 \
    --model_dir=efficientdet-d1-sgd_anchor \
    --ckpt=efficientdet-d1 \
    --train_batch_size=22 \
    --eval_batch_size=1 --eval_samples=1000 \
    --hparams=voc_config.yaml \
[...]

PR #742, opened 2020-09-06
Simplify estimator train_and_eval.

mingxingtan:
After asking TF team, looks like tf.estimator.train_and_eval is recommended for continuous train_and_eval, as it avoids rebuild graph for every train.

### DIFF
--- a/efficientdet/main.py
+++ b/efficientdet/main.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 # ==============================================================================
 """The main training script."""
-import multiprocessing
 import os
 from absl import app
 from absl import flags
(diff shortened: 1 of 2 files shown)


fsx950223:
lgtm

NikZak (commented):
It's definitely more straightforward and it's the correct usage so we should adopt this change. I saw in the issues (#719) that there is memory leak in pure train mode so this change might bring back the memory leak. But we may try solve it from there

mingxingtan:
@NikZak If the OOM still exists, we can add back the multiprocessing option. Could you help give it a test?

efficientdet/main.py
```diff
@@ -254,56 +243,36 @@ def _can_partition(spatial_dim):
...
+      FLAGS.validation_file_pattern,
+      is_training=False,
+      use_fake_data=FLAGS.use_fake_data,
+      max_instances_per_image=max_instances_per_image)
 
   # start train/eval flow.
   if FLAGS.mode == 'train':
-    total_examples = int(config.num_epochs * FLAGS.num_examples_per_epoch)
-    _train(total_examples // FLAGS.train_batch_size)
+    total_examples = int(config.num_epochs * FLAGS.num_examples_per_epoch),
```
fsx950223, line 272: Could I remove the line?

mingxingtan: Oh, yes, this line should be deleted!
I saw you already fixed in the other PR. Thanks for the fix!

PR #747, opened 2020-09-08
added save_checkpoints_steps as flag

sadransh:


fsx950223:
Could you add it to TpuConfig too?

PR #756, opened 2020-09-10
Internal changes

mingxingtan:
PiperOrigin-RevId: 330747985


fsx950223:
LGTM

kartik4949:
@mingxingtan shouldnt we add  test_dataloader in dataloader_test.py which tests dataloader end to end i.e with all augmentation and all

PR #764, opened 2020-09-12
resolved model_dir error when saving 'config.yaml' if directories do not exist before

sadransh:


mingxingtan:
Thanks for the fix.

PR #780, opened 2020-09-17
remove trainable

### LINKED ISSUE #676: No detections after training on custom dataset

fsx950223:
Fix #676 
Only happens when moving average decay = 0 and use_keras_model = True.
Reason: utils.get_ema_variable only return trainable variables that cause WSM variables aren't restored when set is_training_bn = False.

### DIFF
--- a/efficientdet/keras/efficientdet_keras.py
+++ b/efficientdet/keras/efficientdet_keras.py
@@ -124,8 +124,7 @@ class FNode(tf.keras.layers.Layer):
       name = 'WSM' + ('' if i == 0 else '_' + str(i))
       self.vars.append(
           self.add_weight(
-              initializer=initializer, name=name,
-              trainable=self.is_training_bn))
+              initializer=initializer, name=name))
 
   def build(self, feats_shape):
     for i, input_offset in enumerate(self.inputs_offsets):


fsx950223:
I'm confused the fix is correct or I should change
https://github.com/google/automl/blob/00e45119b1fe693c67efb68d98b5c5ddf1786f5a/efficientdet/utils.py#L79
to 
```
ema_vars = tf.variables() + \
```

fsx950223:
I also rewrite a tf2 inference module which fix the bug too.

efficientdet/keras/efficientdet_keras.py
```diff
@@ -124,8 +124,7 @@ def _add_wsm(self, initializer):
       name = 'WSM' + ('' if i == 0 else '_' + str(i))
       self.vars.append(
           self.add_weight(
-              initializer=initializer, name=name,
-              trainable=self.is_training_bn))
```
mingxingtan, line 128: Oops, I believe this trainable should always be true.

mingxingtan (approved):
This is awesome. Thanks for the fix.

PR #788, opened 2020-09-21
add more test cases

fsx950223:


efficientdet/keras/inspector_test.py
```diff
@@ -0,0 +1,73 @@
...
+    if os.path.exists(self.savedmodel_dir):
+      shutil.rmtree(self.savedmodel_dir)
+
+  def tearDown(self):
+    super().tearDown()
+    shutil.rmtree(self.tempdir)
+
+  def test_dry(self):
+    result = subprocess.run(
+        ['python3', 'keras/inspector.py', '--mode=dry', '--ckpt_path=_'])
```
mingxingtan, line 44: Maybe we we could make it simpler by setting flags? such as:

from keras import inspector
from absl import flags
FLAGS = flags.FLAGS


def test_dry(self):
  FLAGS.mode = 'dry'
  FLAGS.ckpt_path = '_'
  inspector.main(None)
  self.assert(...) 

What do you think?

mingxingtan (commented):
Thanks for adding these tests.

efficientdet/keras/inspector_test.py
fsx950223: FLAGS shares values in different test cases which breaks other test cases.
For example: ```infer_test``` gets saved_model_dir which is saved in ```export_test```.

mingxingtan: you can use flagsaver. I will merge this PR, and send you a new PR to show the usage.

efficientdet/keras/inference_test.py
```diff
@@ -24,20 +24,62 @@ class InferenceTest(tf.test.TestCase):
...
+      self.assertEqual(tf.reduce_mean(scores), 0.01000005)
+      self.assertEqual(tf.reduce_mean(classes), 1)
+      self.assertEqual(tf.reduce_mean(valid_lens), 100)
+    elif policy.name == 'float16':
+      pass
+    elif policy.name == 'bfloat16':
+      pass
     self.assertEqual(boxes.shape, (1, 100, 4))
     self.assertEqual(scores.shape, (1, 100))
     self.assertEqual(classes.shape, (1, 100))
```
mingxingtan, line 85: This file doesn't have the tf.test.main. I wonder how did you run this test?

fsx950223: By Pycharm.

PR #789, opened 2020-09-21
Improve handling of custom label maps

glenvorel:
Although the conversion script supports loading custom label map (via the `label_map_json_path` flag), it always tries to load the list of examples from `aeroplane_<set>.txt` file. This fails in most cases (unless the custom dataset actually does contain the `aeroplane` class).

This PR modifies the behavior and instead of using hard-coded class, it picks the class directly from `label_map_dict`.

It also creates the output directory, if it doesn't exist.


glenvorel:
@googlebot I signed it!

mingxingtan (approved):
Thanks for the fix.

PR #797, opened 2020-09-23
Stop if error in child process

NikZak:
Ability to stop the epoch cycle if error in child process. Currently next iteration is started upon OOM


efficientdet/main.py
```diff
@@ -334,22 +334,33 @@ def get_estimator(global_batch_size):
...
+    p_return = multiprocessing.Value('i', 0)
     for e in range(current_epoch + 1, config.num_epochs + 1, epochs_per_cycle):
       if FLAGS.run_epoch_in_child_process:
-        p = multiprocessing.Process(target=run_train_and_eval, args=(e,))
+        p = multiprocessing.Process(
+            target=run_train_and_eval, args=(e, p_return))
         p.start()
         p.join()
+        if p_return.value == 1:
+          break
```
mingxingtan, line 363: How about just return p.exitcode?

mingxingtan: I meant something like:

```
if not p.exitcode:
  return p.exitcode
```

NikZak: Good point. Changed

PR #803, opened 2020-09-24
Fix bugs

### LINKED ISSUE #799: Exported saved_model twice as large as frozen.pb or TensorRT
When I export using:
```bash
python efficientdet/model_inspect.py --runmode=saved_model --model_name=efficientdet-d4 ^
  --ckpt_path=model/efficientdet-d4-scratch ^
  --saved_model_dir=savedmodeldir ^
  --hparams="data/config.yaml"
```
The savedmodeldir have the saved_model.pb and variables.data-00000-of-00001 files but the variables.data-00000-of-00001 file is 158 MB almost twice as large as the frozen.pb or TensorRT files.

Currently I work around that by specifying moving_average_decay: 0 in the config.yaml file when exporting. When I export with --runmode=saved_model and specify moving_average_decay: 0 the variables.data-00000-of-00001 file size reduces to 79.0 MB (and inference seems to be faster or uses less memory. I have not benchmarked this.)

Will it be better to always set moving_average_decay: 0 when using --runmode=saved_model? (or another method may be to remove ema/training variables after the graph is loaded (with --runmode=saved_model) and before export is done?)
Any advice will be appreciated. I also have mentioned this in issue https://github.com/google/automl/issues/676#issuecomment-692364529.

### LINKED ISSUE #801: Metric Learning Rate has disappeared ?
Metric Learning Rate has disappeared in Tensorboard with new code ( last week I think )
It's a pitty it is very usefull

fsx950223:
Fix #801 .
Fix #799 .

### DIFF
--- a/efficientdet/det_model_fn.py
+++ b/efficientdet/det_model_fn.py
@@ -340,7 +340,9 @@ def _model_fn(features, labels, mode, params, model, variable_filter_fn=None):
   Raises:
     RuntimeError: if both ckpt and backbone_ckpt are set.
   """
-  utils.image('input_image', features)
+  is_tpu = params['strategy'] == 'tpu'
+  if params['img_summary_steps']:
+    utils.image('input_image', features, is_tpu)
   training_hooks = []
   params['is_training_bn'] = (mode == tf.estimator.ModeKeys.TRAIN)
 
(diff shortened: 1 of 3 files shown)


mingxingtan (approved):
Looks great. Thanks for the fix!

PR #806, opened 2020-09-25
default gradient_checkpointing_list ['AddN', 'Add_']

NikZak:
This is a better default for gradient_checkpointing_list.

Currently it is ['Add'] and, therefore, BiasAdd nodes are included in checkpoints - this was unintended as the intention was to include only "bottleneck" nodes.

With the new default, the number of checkpoints to include decreases (e.g. for d6 it goes from ~300 to ~100) so it uses slightly less memory. I am not able to measure exact decrease in memory usage as I noticed that tf reserves more memory (all available memory) than it actually needs. But I am able to train more models. E.g. d6 with batch 2 with XLA enabled was giving me OOM and now it trains fine. For some not fully understood reason (maybe saving/reading checkpoints from cache is a more expensive operation than I originally thought) the overall speed of the forward+backward pass not only does not go down but actually goes up. 

There might be even better defaults. I tried some graph vertex cut algorithms (JA-BE-JA-VC) but that did not give a better result thus far

### DIFF
--- a/efficientdet/README.md
+++ b/efficientdet/README.md
@@ -385,7 +385,7 @@ Check these links for a high-level idea of what gradient checkpointing is doing:
 
 **gradient_checkpointing: True**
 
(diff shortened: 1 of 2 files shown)


mingxingtan:
Looks good. Thanks @NikZak

PR #811, opened 2020-09-29
Allow FP16 focal loss compute and fix a few comments.

mingxingtan:


fsx950223 (approved):

NikZak (approved):

efficientdet/det_model_fn.py
```diff
@@ -613,15 +612,15 @@ def scaffold_fn():
...
+    class OomReportingHook(tf.estimator.SessionRunHook):
 
-        def before_run(self, run_context):
-          return tf.estimator.SessionRunArgs(
-              fetches=[],
-              options=tf.RunOptions(report_tensor_allocations_upon_oom=True))
+      def before_run(self, run_context):
+        return tf.estimator.SessionRunArgs(
+            fetches=[],
+            options=tf.RunOptions(report_tensor_allocations_upon_oom=True))
```
fsx950223, line 621: Does oom  tracker will decrease training speed?

mingxingtan: It shouldn't, as it  is triggered only if OOM happens. Did you have this experience before?

fsx950223: I speed up training process by remove profilehook and oomhook.
I'm not care about each hook's performance.

mingxingtan: You are right, this hook indeed slow down the training, I have disabled it in default.

fsx950223 (approved):

PR #833, opened 2020-10-21
add model optimization

fsx950223:
add model optimization example
Screenshot from 2020-10-21 18-53-37
model prune

### DIFF
--- a/efficientdet/keras/train_lib.py
+++ b/efficientdet/keras/train_lib.py
@@ -27,6 +27,167 @@ import utils
 from keras import anchors
 from keras import efficientdet_keras
 
+from tensorflow_model_optimization.python.core.sparsity.keras import pruning_wrapper
+
+
+class MomentumOptimizer(tf.keras.optimizers.Optimizer):
+  def __init__(self, learning_rate, momentum,
+               use_nesterov=False, name='Momentum', **kwargs):
+    super().__init__(name=name, **kwargs)
+    self._set_hyper('learning_rate', learning_rate)
+    self._set_hyper('decay', self._initial_decay)
+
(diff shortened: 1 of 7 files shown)


fsx950223:
I could get 72 mAP0.5 when I train the model with Estimator but I only get 17mAP0.5 when I train the model with Keras on VOC2007. Do you have any idea? @mingxingtan

efficientdet/keras/train_lib.py
```diff
@@ -174,7 +289,7 @@ def get_optimizer(params):
   if params['optimizer'].lower() == 'sgd':
     logging.info('Use SGD optimizer')
     optimizer = tf.keras.optimizers.SGD(
-        learning_rate, momentum=params['momentum'])
+        learning_rate, momentum=params['momentum'], nesterov=True)
```
mingxingtan, line 292: do you get better results with nesterov?

efficientdet/keras/train_lib.py
```diff
@@ -456,9 +584,15 @@ class and box losses from all levels.
               cls_loss, [bs, width, height, -1, self.config.num_classes])
         cls_loss *= tf.cast(
             tf.expand_dims(
-                tf.not_equal(labels['cls_targets_%d' % (level + 3)], -2), -1),
+                tf.not_equal(
+                    labels['cls_targets_%d' % (level + self.config.min_level)],
```
mingxingtan, line 588: Good catch!

mingxingtan (approved):
Very cool feature. Thanks @fsx950223

efficientdet/keras/train_lib.py
fsx950223: I removed nesterov and got 39 mAP0.5

PR #846, opened 2020-10-31
update readme

fsx950223:


efficientdet/keras/train.py
```diff
@@ -140,13 +139,13 @@ def main(_):
       strategy=FLAGS.strategy,
       batch_size=FLAGS.batch_size,
       num_shards=ds_strategy.num_replicas_in_sync)
-
+  config.__dict__.update(params)
```
mingxingtan, line 142: config.override(params)

or 
config.override(params, allow_new_keys=True) if you want allow new members

mingxingtan (approved):
Looks great. Thanks @fsx950223

PR #851, opened 2020-11-04
Release keras model.

fsx950223:


fsx950223:
I got slightly higher mAP than tf1.x which trained on VOC2007. Could you check it?

efficientdet/keras/efficientdet_keras.py
```diff
@@ -26,7 +26,7 @@
 from keras import fpn_configs
 from keras import postprocess
 from keras import util_keras
-from keras import model_optimization
+from keras.model_optimization import get_method as tfmot_method
```
mingxingtan, line 29: the standard format is "from folders import module_name".
My suggestion is to rename model_optimization.py as "tfmot.py", then here you can do "from keras import tfopt", and later you can call: tfopt.get_method(method)

efficientdet/keras/train_lib.py
```diff
@@ -386,6 +387,32 @@ def get_callbacks(params):
...
+
+  def build(self, model, loss_fn, tape):
+    self.model = model
+    self.loss_fn = loss_fn
+    self.tape = tape
+    self.built = True
+
+  def call(self, features, y, y_pred, labeled_loss):
+    return self.adv_config.multiplier * nsl.keras.adversarial_loss(features,
+                               y,
```
mingxingtan, line 409: This library looks cool.

BTW, the indent looks a little strange. Could you also format the code with yapf?

efficientdet/keras/train_lib.py
```diff
@@ -724,7 +751,7 @@ def test_step(self, data):
                                       loss_vals)
       total_loss += det_loss
     if 'segmentation' in self.config.heads:
-      seg_loss_layer = self.loss['seg_loss']
+      seg_loss_layer = self.loss[tf.keras.losses.SparseCategoricalCrossentropy.__name__]
```
mingxingtan, line 754: What's the benefit of using "tf.keras.losses.SparseCategoricalCrossentropy.__name__" instead of a simple "seg_loss"?

mingxingtan (commented):
Overall looks great. Just a few minor comments.

efficientdet/keras/train_lib.py
fsx950223: 1. Replace the magic value
2. I could tape into definition when I browse the code.

efficientdet/keras/efficientdet_keras.py
fsx950223: I try to ignore import errors when I import the module on the environment which didn't install tensorflow_model_optimization.
Any better solution?

efficientdet/det_model_fn.py
```diff
@@ -413,23 +379,8 @@ def model_fn(inputs):
...
-        return grad.gradients(
-            ys,
-            xs,
-            grad_ys,
-            checkpoints=params['device']['grad_ckpting_list'],
-            **kwargs)
-
-      gradients.__dict__['gradients'] = gradients_
+    elif params['mixed_precision']:
+      optimizer = tf.train.experimental.enable_mixed_precision_graph_rewrite(optimizer)
```
fsx950223, line 383: Should gain better performance and less bugs.
Fix https://github.com/google/automl/issues/852

mingxingtan (approved):

PR #854, opened 2020-11-06
Create python-app.yml

fsx950223:


efficientdet/install_deps.sh
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+for line in $(cat efficientdet/requirements.txt)
+do
+  pip install $line
+done
```
juliangrosshauser, line 20: Wouldn't the usual `pip install --requirement efficientdet/requirements.txt` be easier here?

fsx950223: I want to install python packages in requirements.txt order.

PR #857, opened 2020-11-09
Sync changes: mostly with format changes

mingxingtan:
PiperOrigin-RevId: 341344350

### DIFF
--- a/efficientdet/keras/train_lib.py
+++ b/efficientdet/keras/train_lib.py
@@ -18,23 +18,22 @@ import math
 import os
 import re
 from absl import logging
+import neural_structured_learning as nsl
 import numpy as np
-import tensorflow as tf
 
+import tensorflow as tf
 import inference
 import iou_utils
 import utils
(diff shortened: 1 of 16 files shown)


efficientdet/keras/train_lib.py
```diff
@@ -488,12 +488,12 @@ def __init__(self, delta=0.1, **kwargs):
   def call(self, y_true, box_outputs):
     num_positives, box_targets = y_true
     normalizer = num_positives * 4.0
-    mask = tf.cast(box_targets != 0.0, num_positives.dtype)
+    mask = tf.cast(box_targets != 0.0, box_targets.dtype)
```
mingxingtan, line 491: num_positives could be a scalar (has no .dtype), so I changed it to box_targets.

fsx950223: I changed it in another PR too.

fsx950223 (approved):

PR #859, opened 2020-11-11
Reset NMS threshold to default, fix accuracy regression

ngaloppo:


fsx950223:
Could you set score threshold to 0 before [eval](https://github.com/google/automl/blob/4dfd77b6526edc0eda1066c2ca8e57ed0a135a92/efficientdet/det_model_fn.py#L425)

mingxingtan:
Looks like @fsx950223 has already changed the default value, but thank you for the fix, @ngaloppo

mingxingtan:
I am going to close this PR.

Issue #838, opened 2020-10-27
Visualize random images from dataloader


kartik4949:
Lot of people are having issue and are skeptical if the tfrecords created are correct or data labelling is correct 
Im developing a feature in dataloader which visualizes random 'n' number of images before training or just for testing dataloader!
@mingxingtan @fsx950223 thoughts?

fsx950223:
https://github.com/sulc/tfrecord-viewer

kartik4949:
@fsx950223 so cant we add this or something like this in our repository!

Steve-2040:
Sounds like a good idea. I think it will be helpful if the output-sample-diagnostic-images also include augmented images? So, this will need to be done while training? Not sure if this can be done for GPU and TPU training.

kartik4949:
@Steve-2040 yes will make a flag which can be invoked to diagnose training images

Steve-2040:
I think this can be helpful. If you can specify a diagnostics path and folder (directory) to save 1 in n samples. It can be a guessing game to know what images are actually used for training.

kartik4949:
Exactly 
Will store some n sample images with rois or masks before actuall training starts and user can confirm the images and continue with training

fsx950223:
> I think this can be helpful. If you can specify a diagnostics path and folder (directory) to save 1 in n samples. It can be a guessing game to know what images are actually used for training.

Why don't you use ```img_summary_steps```?

kartik4949:
@fsx950223 in tensorboard ...
Yeah that can be a option too but i didnt see any image step in tensorboard

Steve-2040:
@fsx950223 Thanks for mentioning that. I was not aware of img_summary_steps Does that output the visual images including augmentation?
@kartik4949 I am thinking even if it starts outputting images while training, the user can verify these images after training started and then decide to continue or restart training.

kartik4949:
use this #865

PR #865, opened 2020-11-15
:zap: Added visualization script

kartik4949:
Solves #838 
I think this will be helpful for early signs of bad dataset, it can save unnecessary issues and debugging time which could have been a silly data format or labelling issue.
I can add this script to visualize folder too.
@mingxingtan @fsx950223 thoughts?

p.s: @fsx950223 suggested to use tfrecords viewer but I think this solves third party tool usages and will be very handy.

### DIFF
--- a/efficientdet/README.md
+++ b/efficientdet/README.md
@@ -390,4 +390,29 @@ Testing shows that:
 * It allows to train a d7x network with batch size of 2 by keras/train.py on a 11Gb (1080Ti) GPU
 * It also allows to train a d6 network with batch size of 2 by main.py on a 11Gb (1080Ti) GPU
 
+## 12. Visualize TF-Records.
+
+You can visualize tf-records with following commands:
+
+To visualize training tfrecords with input dataloader use.
+```
+python dataset/inspect_tfrecords.py --file_pattern dataset/sample.record\ 
+--model_name "efficientdet-d0" --samples 10\ 
+--save_samples_dir train_samples/  -hparams="label_map={1:'label1'}, autoaugmentation_policy=v3"
+
+```
+
+To visualize evaluation tfrecords use.
+```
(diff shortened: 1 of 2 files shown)


kartik4949:
@mingxingtan if it is a go from you, i can proceed with this pr and complete it. i.e configurable, adding label, readme update.

mingxingtan:
Hi @kartik4949  Thanks for adding this. Please don't modify TF1 version of main.py as we are moving to TF2.
Looks like this PR is only about dataset, maybe you can add a standalone tool in dataset (such as dataset/inspect_tfrecord.py)?

kartik4949:
@mingxingtan alrite, makes sense
Let me add this in dataset
Thanks!

kartik4949:
@mingxingtan  done, can you suggest tests?

kartik4949:
@mingxingtan 
so this script saves  images, bboxes formed by input dataloader with given flag '-train', configuration passed in config.yaml /hparams.

Visualize train tfrecord with augmentation.
``` 
python dataset/inspect_tfrecords.py --file_pattern dataset/sample.record --model_name "efficientdet-d0" --save_samples_dir visualize_dataset/  -train -hparams="label_map={1:'type1'}, autoaugment_policy=v3"

# above command builds dataloader equivalent with config.yaml/hparams.
# visualizes images and bboxes with given label_map.
# train = True
```

mingxingtan (approved):
LGTM. Thanks for adding this tool

ngaloppo:
@kartik4949 What's the difference between train & eval mode for the visualization tool? Why do you need both?

kartik4949:
@ngaloppo eval mode wont have any augmentations 
Train will have it

kartik4949:
a

PR #868, opened 2020-11-16
add tflite infer script

fsx950223:


mingxingtan:
Excellent. Thanks @fsx950223

Issue #872, opened 2020-11-17
Shouldnt we normalize our images in inference too.


kartik4949:
Shouldnt we normalize our image in inference too, like we did in the training process ('main.py'), 
```
def normalize_image(self):
    """Normalize the image to zero mean and unit variance."""
    # The image normalization is identical to Cloud TPU ResNet.
    self._image = tf.image.convert_image_dtype(self._image, dtype=tf.float32)
    offset = tf.constant([0.485, 0.456, 0.406])
    offset = tf.expand_dims(offset, axis=0)
    offset = tf.expand_dims(offset, axis=0)
    self._image -= offset

    scale = tf.constant([0.229, 0.224, 0.225])
    scale = tf.expand_dims(scale, axis=0)
    scale = tf.expand_dims(scale, axis=0)
    self._image /= scale

```

My question is shouldn't we do this normalization for inference??
@mingxingtan @fsx950223 
if  I'm wrong , please let me know how we are doing this:(

PR #873, opened 2020-11-17
inference normalization fix

kartik4949:
Solves #872 
Please test on sample image. @mingxingtan @fsx950223


mingxingtan:
The image preprocessing is done here: https://github.com/google/automl/blob/9af8a40169b0f5f07c61a3e09ab0077f1f476fdc/efficientdet/keras/efficientdet_keras.py#L894

I don't think you need to normalize images by yourself.

kartik4949:
> The image preprocessing is done here: https://github.com/google/automl/blob/9af8a40169b0f5f07c61a3e09ab0077f1f476fdc/efficientdet/keras/efficientdet_keras.py#L894
> 
> 
> 
> I don't think you need to normalize images by yourself.

Thanks got it

Issue #839, opened 2020-10-27
Warmup init learning rate is not adjusted


rbournhonesque:
The optimizer learning rate is adjusted with respect to the batch size:

```
  params['adjusted_learning_rate'] = (
      params['learning_rate'] * batch_size / _DEFAULT_BATCH_SIZE)
```

However, `lr_warmup_init` is not adjusted, which is not the desired behaviour.
In my trainings, the "warmup" stage was actually an *increase* of the learning rate from `lr_warmup_init` to `adjusted_learning_rate`.

staeff777:
I agree.
```
params['batch_size'] = 4
params['num_epochs'] = 25
params['num_shards'] = 1
params['num_examples_per_epoch'] = 1000
params['learning_rate']= 0.01
params['lr_warmup_init']=0.001
params['lr_warmup_epoch']=1
```
results in 
```
adjusted_learning_rate': 0.000625
```
that produces following cosine curve whith no warmup:
![grafik](https://user-images.githubusercontent.com/534957/99377370-69e23480-28c6-11eb-8719-8a90dc4cc8d5.png)

PR #874, opened 2020-11-17
Adjust lr warmup init to batch size (#839)

rbournhonesque:
Fix issue #839


rbournhonesque:
@googlebot I signed it!

mingxingtan:
Thanks for the fix.

PR #876, opened 2020-11-18
Fix tf nms bug

fsx950223:


efficientdet/keras/postprocess.py
```diff
@@ -176,7 +176,7 @@ def nms(params, boxes: T, scores: T, classes: T,
     score_thresh = nms_configs['score_thresh'] or float('-inf')
   elif method == 'gaussian':
     sigma = nms_configs['sigma'] or 0.5
-    iou_thresh = nms_configs['iou_thresh'] or 0.3
+    iou_thresh = 1.0
```
fsx950223, line 179: Since https://github.com/tensorflow/tensorflow/blob/8a13606bbebf13d07083e84a1d640a5e7339f8c9/tensorflow/core/kernels/image/non_max_suppression_op.cc#L199 is selected by iou_thresh, change it to 1.0 in order to avoid weight suppress.

mingxingtan: Oh good catch. Should it be "sim <= similarity_threshold ? static_cast<T>(1.0): weight;"?
Could you help create a PR? I will ping TF owners to help review it. Thanks!

mingxingtan (approved):

PR #878, opened 2020-11-18
Packaged application

madisi98:
Added setup.py that allows the installation of the package, making working with it a lot easier.
Also added entry points for every script that's used on both keras and tensorflow README.md

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,43 @@
+import setuptools
+
+
+setuptools.setup(
+    name             = 'efficientdet',
+    version          = '0.0.0',
+    description      = '',
+    url              = '',
+    author           = 'PyPackager',
+    author_email     = '',
+    maintainer       = '',
+    maintainer_email = '',
+    packages         = setuptools.find_packages(),
+    install_requires = [
+        'lxml>=4.6.1',
+        'absl-py>=0.7.1',
+        'matplotlib>=3.0.3',
+        'numpy>=1.16.4',
+        'Pillow>=6.0.0',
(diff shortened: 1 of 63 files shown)


madisi98:
@googlebot I signed it!

setup.py
```diff
@@ -0,0 +1,43 @@
...
+        'PyYAML>=5.1',
+        'six>=1.12.0',
+        'tensorflow>=2.3.0',
+        'tensorflow-addons>=0.11.2',
+        'neural-structured-learning>=1.3.1',
+        'tensorflow-model-optimization>=0.5',
+        'Cython>=0.29.13',
+    ],
+    dependency_links = [
+        'git+https://github.com/cocodataset/cocoapi.git#subdirectory=PythonAPI'
```
juliangrosshauser, line 29: https://pypi.org/project/pycocotools is a lot easier to work with, because it's simply the packaged version of the above git link. It's also used in Detectron2 (the maintainer of the pycocotools package on PyPI is also a Detectron2 maintainer): https://github.com/facebookresearch/detectron2/blob/master/setup.py#L202

madisi98: I prefer that to, but decided to leave the one that is bacause it's what appears in the requierements.txt

juliangrosshauser: I thought I mention it, because this PR might be a great opportunity to switch.

madisi98: Okay, I have to fix the tests too so I will switch to draft and publish the PR when those 2 things are done.

PR #888, opened 2020-11-27
Sync Internal changes

mingxingtan:
PiperOrigin-RevId: 344144225

### DIFF
--- a/efficientdet/dataset/inspect_tfrecords.py
+++ b/efficientdet/dataset/inspect_tfrecords.py
@@ -12,17 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-"""Inspection of dataset"""
-
+"""Inspect tfrecord dataset."""
 import os
-import sys
-
+from absl import app
+from absl import flags
+from absl import logging
 import numpy as np
-import tensorflow as tf
-from absl import app, flags, logging
 from PIL import Image
+import tensorflow as tf
 
-sys.path.append('./')
 import dataloader
 import hparams_config
 import utils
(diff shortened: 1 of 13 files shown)


efficientdet/keras/train_lib.py
```diff
@@ -733,11 +739,21 @@ def train_step(self, data):
         tf.summary.image('input_image', images)
     with tf.GradientTape() as tape:
       if len(self.config.heads) == 2:
-        cls_outputs, box_outputs, seg_outputs = self(images, training=True)
+        cls_outputs, box_outputs, seg_outputs = util_keras.fp16_to_fp32_nested(
```
mingxingtan, line 742: It is better to use float32 for loss computing.

mingxingtan (commented):
This PR is mostly about simple format changes.

fsx950223 (approved):

efficientdet/dataset/inspect_tfrecords.py
```diff
@@ -81,7 +77,7 @@ def visualize(self):
     scale_image = tf.constant([0.229, 0.224, 0.225])
     scale_image = tf.reshape(scale_image, (1, 1, -1))
 
-    logging.info(f"Visualizing TfRecords {FLAGS.file_pattern} ....")
+    logging.info('Visualizing TfRecords %s', FLAGS.file_pattern)
```
kartik4949, line 80: @mingxingtan i thought using f strings is better way as of python >=3.6

mingxingtan: Hi @kartik4949, yes, f-strings are generally good, but logging prefers %s for the benefit of lazy evaluation.

f-string would cause pylint warning logging-format-interpolation, see: https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation?rq=1

kartik4949: Informative :) thanks

PR #895, opened 2020-12-03
Fix mirrored variable load

fsx950223:


efficientdet/utils.py
```diff
@@ -111,35 +111,47 @@ def get_ckpt_var_map(ckpt_path, ckpt_scope, var_scope, skip_mismatch=None):
...
+      tf.distribute.get_replica_context().replica_id_in_sync_group)
+  else:
+    replica_id = 0
+
   for i, v in enumerate(model_vars):
-    if not v.op.name.startswith(var_scope):
+    var_op_name = v.op.name
+
+    if replica_id >= 1:
+      var_op_name = ''.join(var_op_name.rsplit(f'/replica_{replica_id}', 1))
```
mingxingtan, line 124: Good idea! Thanks @fsx950223

mingxingtan: For mirror strategy and replica_x, does x start from 0 or 1?

fsx950223: replica_x starts at 1.
For the host, there is no replica_x in variable.op.name

PR #905, opened 2020-12-14
fix label smooth

fsx950223:


fsx950223:
It's a sigmoid loss, my mistake.

PR #908, opened 2020-12-15
update for tf2.4

fsx950223:
cc @mingxingtan 
I met some issues when I train the model with keras multi gpus on tf2.4, could you check it?
I wonder it's a bug of my environment or tensorflow.

### DIFF
--- a/efficientdet/dataloader.py
+++ b/efficientdet/dataloader.py
@@ -48,6 +48,14 @@ class InputProcessor:
     self._crop_offset_y = tf.constant(0)
     self._crop_offset_x = tf.constant(0)
 
+  @property
+  def image(self):
+    return self._image
+
+  @image.setter
+  def image(self, image):
+    self._image = image
+
   def normalize_image(self):
     """Normalize the image to zero mean and unit variance."""
     # The image normalization is identical to Cloud TPU ResNet.
@@ -61,6 +69,7 @@ class InputProcessor:
     scale = tf.expand_dims(scale, axis=0)
     scale = tf.expand_dims(scale, axis=0)
     self._image /= scale
+    return self._image
 
   def set_training_random_scale_factors(self,
                                         scale_min,
@@ -126,6 +135,7 @@ class InputProcessor:
 
   def resize_and_crop_image(self, method=tf.image.ResizeMethod.BILINEAR):
     """Resize input image and crop it to the self._output dimension."""
+    dtype = self._image.dtype
     scaled_image = tf.image.resize(
         self._image, [self._scaled_height, self._scaled_width], method=method)
     scaled_image = scaled_image[self._crop_offset_y:self._crop_offset_y +
@@ -135,7 +145,8 @@ class InputProcessor:
     output_image = tf.image.pad_to_bounding_box(scaled_image, 0, 0,
                                                 self._output_size[0],
                                                 self._output_size[1])
-    return output_image
+    self._image = tf.cast(output_image, dtype)
+    return self._image
 
 
 class DetectionInputProcessor(InputProcessor):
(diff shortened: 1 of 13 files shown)


efficientdet/keras/train.py
```diff
@@ -74,7 +74,7 @@
 flags.DEFINE_integer('batch_size', 64, 'training batch size')
 flags.DEFINE_integer('eval_samples', 5000, 'The number of samples for '
                      'evaluation.')
-flags.DEFINE_integer('steps_per_execution', 1000,
+flags.DEFINE_integer('steps_per_execution', 1,
```
fsx950223, line 77: Disable it in default since there is some issues in multi gpus training with uninitlized optimizer.

efficientdet/dataloader.py
```diff
@@ -294,6 +305,21 @@ def dataset_parser(self, value, example_decoder, anchor_labeler, params):
...
+          input_processor.random_horizontal_flip()
+
+        input_processor.set_training_random_scale_factors(
+            params['jitter_min'], params['jitter_max'],
+            params.get('target_size', None))
+      else:
+        input_processor.set_scale_factors_to_output_size()
+
+      image = input_processor.resize_and_crop_image()
+      boxes, classes = input_processor.resize_and_crop_boxes()
```
fsx950223, line 321: Resize image first could double speed up pipeline.

mingxingtan: Interesting finding!

efficientdet/utils.py
```diff
@@ -254,9 +254,7 @@ def batch_norm_class(is_training, strategy=None):
   if is_training and strategy == 'tpu':
     return TpuBatchNormalization
   elif is_training and strategy == 'gpus':
-    # TODO(fsx950223): use SyncBatchNorm after TF bug is fixed (incorrect nccl
-    # all_reduce). See https://github.com/tensorflow/tensorflow/issues/41980
-    return BatchNormalization
+    return SyncBatchNormalization
```
mingxingtan, line 257: How about the speed of SyncBatchNormalization for multiple GPUs?

mingxingtan (approved):

efficientdet/utils.py
```diff
@@ -622,13 +622,11 @@ def build_model_with_precision(pp, mm, ii, *args, **kwargs):
     inputs = tf.cast(ii, tf.bfloat16)
     with tf.tpu.bfloat16_scope():
       outputs = mm(inputs, *args, **kwargs)
-    set_precision_policy('float32')
```
fsx950223, line 632: After remove 2 lines, I could train model with recompute_grad and mixed_precision.
Why set policy back to float32? Could I remove them? @mingxingtan

efficientdet/utils.py
fsx950223: It's slower about 40% than BatchNormalization for multiple GPUs. I believe it's acceptable.

fsx950223: According to https://github.com/tensorflow/tensorflow/blob/9489702e35b16a40a1accf3b8b5ed557efae10c7/tensorflow/python/keras/layers/normalization_v2.py#L151.
Should I split replica_ctx.all_reduce?

efficientdet/utils.py
mingxingtan: Yes, please feel free to remove it. it is not necessary.

efficientdet/keras/train.py
mingxingtan: sgtm.

efficientdet/utils.py
mingxingtan: I did not quite understand these comments, but I think 40% slower is fine.

mingxingtan (approved):

PR #909, opened 2020-12-20
Sync internal code changes of tflite/tfhub.

mingxingtan:
Mostly contributed by Yiqi Li (yuqili@google.com)


efficientdet/dataloader_test.py
```diff
@@ -13,63 +13,19 @@
...
 """Data loader and processing test cases."""
-import os
-import tempfile
+
 import tensorflow as tf
 
 import dataloader
 import hparams_config
-from dataset import tfrecord_util
+from brain_automl.efficientdet import test_util
```
fsx950223, line 21: test_util

efficientdet/keras/inference_test.py
```diff
@@ -17,6 +17,7 @@
 import tempfile
 from absl import logging
 import tensorflow as tf
+from brain_automl.efficientdet import test_util
```
fsx950223, line 20: import test_util

efficientdet/dataloader_test.py
mingxingtan: Good point. Done.

efficientdet/keras/inference_test.py
mingxingtan: Done

mingxingtan:
@ ziyeqinghan fyi

fsx950223:
Add tensorflow_hub dependency.

efficientdet/keras/train_lib.py
```diff
@@ -843,3 +849,61 @@ def test_step(self, data):
...
+    # class/box output prediction network.
+    num_anchors = len(config.aspect_ratios) * config.num_scales
+
+    if config.separable_conv:
+      conv2d_layer = functools.partial(
+          tf.keras.layers.SeparableConv2D, depth_multiplier=1)
+    else:
+      conv2d_layer = tf.keras.layers.Conv2D
+    self.classes = conv2d_layer(
+        config.num_classes * num_anchors,
```
fsx950223, line 872: Initializers are different from original one.

fsx950223 (approved):

efficientdet/keras/train_lib.py
mingxingtan: It seems to be the same as before: https://github.com/google/automl/blob/890eaddc0963701e515de06a11a662d1f1027770/efficientdet/keras/efficientdet_keras.py#L426

Notably, the last convnet for class prediction is special in order to enable focal loss.

fsx950223: ```pointwise_initializer``` and ```depthwise_initializer``` are ```tf.initializers.variance_scaling()``` in ClassNet, should it be changed? It's better to use ClassNet and BoxNet whose are defined in efficientnet_keras.py

mingxingtan: Good point. I am going to merge this for now, and update it later.

fsx950223 (approved):

PR #960, opened 2021-03-06
Merge internal changes.

mingxingtan:
These changes are largely done by ziyeqinghan@ (Yuqi Li) for TFLite.

### DIFF
--- a/efficientdet/dataloader.py
+++ b/efficientdet/dataloader.py
@@ -118,7 +118,7 @@ class InputProcessor:
 
   def resize_and_crop_image(self, method=tf.image.ResizeMethod.BILINEAR):
     """Resize input image and crop it to the self._output dimension."""
-    dtype = self._image.dtyp
+    dtype = self._image.dtype
     scaled_image = tf.image.resize(
         self._image, [self._scaled_height, self._scaled_width], method=method)
     scaled_image = scaled_image[self._crop_offset_y:self._crop_offset_y +
--- a/efficientdet/dataloader.py
+++ b/efficientdet/dataloader.py
@@ -48,11 +48,21 @@ class InputProcessor:
     self._crop_offset_y = tf.constant(0)
     self._crop_offset_x = tf.constant(0)
 
+  @property
+  def image(self):
+    return self._image
+
+  @image.setter
+  def image(self, image):
+    self._image = image
+
   def normalize_image(self, mean_rgb, stddev_rgb):
     """Normalize the image to zero mean and unit variance."""
+    # The image normalization is identical to Cloud TPU ResNet.
     self._image = tf.cast(self._image, dtype=tf.float32)
     self._image -= tf.constant(mean_rgb, shape=(1, 1, 3), dtype=tf.float32)
     self._image /= tf.constant(stddev_rgb, shape=(1, 1, 3), dtype=tf.float32)
+    return self._image
 
   def set_training_random_scale_factors(self,
                                         scale_min,
--- a/efficientdet/dataloader.py
+++ b/efficientdet/dataloader.py
@@ -48,28 +48,11 @@ class InputProcessor:
     self._crop_offset_y = tf.constant(0)
     self._crop_offset_x = tf.constant(0)
 
-  @property
-  def image(self):
-    return self._image
-
-  @image.setter
-  def image(self, image):
-    self._image = image
-
-  def normalize_image(self):
+  def normalize_image(self, mean_rgb, stddev_rgb):
     """Normalize the image to zero mean and unit variance."""
-    # The image normalization is identical to Cloud TPU ResNet.
-    self._image = tf.image.convert_image_dtype(self._image, dtype=tf.float32)
-    offset = tf.constant([0.485, 0.456, 0.406])
-    offset = tf.expand_dims(offset, axis=0)
-    offset = tf.expand_dims(offset, axis=0)
-    self._image -= offset
-
-    scale = tf.constant([0.229, 0.224, 0.225])
-    scale = tf.expand_dims(scale, axis=0)
-    scale = tf.expand_dims(scale, axis=0)
-    self._image /= scale
-    return self._image
+    self._image = tf.cast(self._image, dtype=tf.float32)
+    self._image -= tf.constant(mean_rgb, shape=(1, 1, 3), dtype=tf.float32)
+    self._image /= tf.constant(stddev_rgb, shape=(1, 1, 3), dtype=tf.float32)
 
   def set_training_random_scale_factors(self,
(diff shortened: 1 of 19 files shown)


efficientdet/keras/util_keras.py
```diff
@@ -180,14 +224,18 @@ def fp16_to_fp32_nested(input_nested):
...
-      kwargs['name'] = 'tpu_batch_normalization'
-    return bn_class(*args, **kwargs)
-  return _wrapper
-
-if tf.compat.v1.executing_eagerly_outside_functions():
-  utils.BatchNormalization = get_batch_norm(tf.keras.layers.BatchNormalization)
-  utils.SyncBatchNormalization = get_batch_norm(tf.keras.layers.experimental.SyncBatchNormalization)
-  utils.TpuBatchNormalization = get_batch_norm(tf.keras.layers.experimental.SyncBatchNormalization)
\ No newline at end of file
+### The following code breaks COCO training.
```
mingxingtan, line 227: @fsx950223  The following code breaks COCO training.

fsx950223: What's problem, bad performance or runtime error.

efficientdet/dataloader.py
```diff
@@ -48,28 +48,11 @@ def __init__(self, image, output_size):
     self._crop_offset_y = tf.constant(0)
     self._crop_offset_x = tf.constant(0)
 
-  @property
-  def image(self):
-    return self._image
-
-  @image.setter
-  def image(self, image):
-    self._image = image
```
fsx950223, line 57: Could keep the api since I use them in other models.

efficientdet/keras/postprocess.py
```diff
@@ -24,13 +24,19 @@
...
+# TFLite regular NMS == postprocess_per_class
+TFLITE_USE_REGULAR_NMS = False
 
 
 def to_list(inputs):
   if isinstance(inputs, dict):
-    return [tf.cast(inputs[k], tf.float32) for k in sorted(inputs.keys())]
+    return [inputs[k] for k in sorted(inputs.keys())]
   if isinstance(inputs, list):
-    return [tf.cast(i, tf.float32) for i in inputs]
```
fsx950223, line 33: I remember it fixs mixed precision eval .

efficientdet/dataloader.py
mingxingtan: Done

efficientdet/keras/postprocess.py
mingxingtan: Done.

efficientdet/keras/efficientdet_keras.py
```diff
@@ -149,10 +148,10 @@ def build(self, feats_shape):
       self._add_wsm('ones')
     elif self.weight_method == 'channel_attn':
       num_filters = int(self.fpn_num_filters)
-      self._add_wsm(lambda shape, dtype, **kwargs: tf.ones(shape, dtype, **kwargs), num_filters)
+      self._add_wsm(lambda: tf.ones([num_filters]))
     elif self.weight_method == 'channel_fastattn':
       num_filters = int(self.fpn_num_filters)
-      self._add_wsm(lambda shape, dtype, **kwargs: tf.ones(shape, dtype, **kwargs), num_filters)
```
fsx950223, line 155: Does it work on tf2?

efficientdet/keras/efficientdet_keras.py
```diff
@@ -932,12 +955,22 @@ def _postprocess(self, cls_outputs, box_outputs, scales, mode='global'):
     if not mode:
       return cls_outputs, box_outputs
 
+    # TODO(tanmingxing): remove this cast once FP16 works postprocessing.
+    cls_outputs = [tf.cast(i, tf.float32) for i in cls_outputs]
```
fsx950223, line 959: It could be removed since casts are done in to_list.

efficientdet/keras/postprocess.py
fsx950223: ```# TODO(tanmingxing): remove this cast once FP16 works postprocessing.```
The comment should be added here?

fsx950223 (approved):
LGTM but some issues.

efficientdet/keras/efficientdet_keras.py
mingxingtan: Good catch. I forgot to merge this change. Updated now.

efficientdet/keras/efficientdet_keras.py
mingxingtan: I feel the type conversion should happen here, instead of to_list.
I didn't into much detail for this part, but what's the reason to remove these casts, and do the cast in to_list?

efficientdet/keras/postprocess.py
mingxingtan: I think if we do the conversion right after the network, then we don't need these casts any more.

fsx950223: https://github.com/google/automl/blob/539ab6554c4b47e319a43da29125714bb4adfe35/efficientdet/keras/train_lib.py#L339
If so, please add cast here too.

mingxingtan: Done.

efficientdet/keras/postprocess.py
```diff
@@ -238,6 +243,133 @@ def postprocess_combined(params, cls_outputs, box_outputs, image_scales=None):
...
+  @tf.function(experimental_implements=tflite_nms_implements_signature(params))
+  # pylint: disable=g-unused-argument,unused-argument
+  def dummy_post_processing(box_encodings, class_predictions, anchor_boxes):
+    boxes = tf.constant(0.0, dtype=tf.float32, name='boxes')
+    scores = tf.constant(0.0, dtype=tf.float32, name='scores')
+    classes = tf.constant(0.0, dtype=tf.float32, name='classes')
+    num_detections = tf.constant(0.0, dtype=tf.float32, name='num_detections')
+    return boxes, classes, scores, num_detections
+
+  return dummy_post_processing(box_outputs, scores, decoded_anchors)[::-1]
```
fsx950223, line 370: Why reverse results?

mingxingtan: Sorry, what do you mean?

fsx950223: [::-1] reverses the results.
Is it necessary?

mingxingtan: Ah, good point. This is just a placeholder, the reverse might not be necessary.
Since tflite team added this, I will double check with them later.

fsx950223 (approved):

PR #976, opened 2021-04-14
Fix inspectors

glenvorel:
Benchmarking with `only_network` was not possible due to incorrect data type of the synthetic tensor.

The `inspect_tfrecords.py` utility ignored all bounding boxes with `ymin` equal to 0. These are now correctly rendered.

It also used red color for all bounding boxes which sometimes made it difficult to distinguish different object classes if they were clustered. It now behaves identically to inference of `keras.inspector` module and uses different color for each class.

Video exported using the `keras.inspector` module now dynamically sets the frame rate of output file equal to the input file.

### DIFF
--- a/efficientdet/keras/inspector.py
+++ b/efficientdet/keras/inspector.py
@@ -103,8 +103,7 @@ def main(_):
     if FLAGS.saved_model_dir:
       driver.load(FLAGS.saved_model_dir)
       if FLAGS.saved_model_dir.endswith('.tflite'):
-        image_size = utils.parse_image_size(model_config.image_size)
-        image_arrays = tf.image.resize_with_pad(image_arrays, *image_size)
+        image_arrays = tf.image.resize_with_pad(image_arrays, *model_config.image_size)
         image_arrays = tf.cast(image_arrays, tf.uint8)
     detections_bs = driver.serve(image_arrays)
     boxes, scores, classes, _ = tf.nest.map_structure(np.array, detections_bs)
@@ -135,8 +134,9 @@ def main(_):
       # use synthetic data if no image is provided.
       image_arrays = tf.ones((batch_size, *model_config.image_size, 3),
                              dtype=tf.uint8)
-      if FLAGS.only_network:
-        image_arrays = tf.cast(image_arrays, tf.float32)
+    if FLAGS.only_network:
+      image_arrays = tf.image.convert_image_dtype(image_arrays, tf.float32)
+      image_arrays = tf.image.resize(image_arrays, model_config.image_size)
     driver.benchmark(image_arrays, FLAGS.bm_runs, FLAGS.trace_filename)
   elif FLAGS.mode == 'dry':
     # transfer to tf2 format ckpt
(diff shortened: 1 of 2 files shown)


kartik4949:
nice, was thinking to use the prebuilt visualizer too, now it makes it more uniform.

kartik4949 (approved):

mingxingtan (approved):
Thanks!

efficientdet/keras/inspector.py
```diff
@@ -135,6 +135,8 @@ def main(_):
       # use synthetic data if no image is provided.
       image_arrays = tf.ones((batch_size, *model_config.image_size, 3),
                              dtype=tf.uint8)
+      if FLAGS.only_network:
+        image_arrays = tf.cast(image_arrays, tf.float32)
```
fsx950223, line 139: ```image_arrays``` should between [0,1].

glenvorel: Hi @fsx950223, thank you for your comment.
1. `Policy` object doesn't have `dtype` attribute, it only has `name`, `variable_dtype` and `compute_dtype`.
And is it really necessary to use global policy when the `ServingDriver` _always_ uses `float32` dtype when `only_network` is enabled? ([inference.py:313](https://github.com/google/automl/blob/f036f3cee588fee9ad93d01e8dc3db901eb95a85/efficientdet/keras/inference.py#L313))
2. `image_arrays` are in range [0, 255] only if `input_image` is provided. And unless the image exactly matches dimensions of the input tensor, benchmarking with `only_network` is not possible. The image would have to be resized. And the `resize` method always returns `float32` dtype - it solves both dimensions and dtype issues.

How about this?
```python
    if FLAGS.only_network:
      image_arrays = tf.image.resize(image_arrays, [*model_config.image_size])
    driver.benchmark(image_arrays, FLAGS.bm_runs, FLAGS.trace_filename)
```

fsx950223:
``` python
    if FLAGS.only_network:
      image_arrays = tf.image.convert_image_dtype(image_arrays,
                                                  tf.keras.mixed_precision.global_policy().compute_dtype)
      image_arrays = tf.image.resize(image_arrays, [*model_config.image_size])
    driver.benchmark(image_arrays, FLAGS.bm_runs, FLAGS.trace_filename)
```

glenvorel: 1. Even if `convert_image_dtype` yielded different dtype than `float32`, then `resize` method would immediately convert it to `float32` anyway. Is this step necessary?
2. Image size is already parsed on line 80, why parsing it again?

fsx950223: Good, remove https://github.com/google/automl/blob/9bff7529b26efd5ef51b574995bdc60530c85ed5/efficientdet/keras/inspector.py#L106 too, and 
```
    if FLAGS.only_network:
      image_arrays = tf.image.convert_image_dtype(image_arrays, tf.float32)
      image_arrays = tf.image.resize(image_arrays, utils.parse_image_size(model_config.image_size))
    driver.benchmark(image_arrays, FLAGS.bm_runs, FLAGS.trace_filename)
```

glenvorel: Done, thank you for your suggestions.

PR #1003, opened 2021-05-17
optimize training speed

fsx950223:
cc @mingxingtan.

### DIFF
--- a/efficientnetv2/main.py
+++ b/efficientnetv2/main.py
@@ -17,6 +17,7 @@ import copy
 import os
 import re
 import time
+import functools
 from absl import app
 from absl import flags
 from absl import logging
@@ -467,7 +468,7 @@ def main(unused_argv):
           ibase = config.data.ibase or (input_image_size / 2)
           isize_list = np.linspace(ibase, input_image_size, total_stages).astype(np.int32)
           ds = tf.data.experimental.Counter(dtype=tf.int32)
-          ds_iters = []
+          fns = {}
           for index, image_size in enumerate(isize_list):
             if config.train.sched:
               config.data.ram = ram_list[index]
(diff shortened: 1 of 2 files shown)


efficientnetv2/main.py
```diff
@@ -464,35 +464,25 @@ def main(unused_argv):
...
+                config.data.ram = ram_list[index]
+                config.data.mixup_alpha = mixup_list[index]
+                config.data.cutmix_alpha = cutmix_list[index]
+              ds_lab_cls = datasets.build_dataset_input(True, image_size,
+                                                        image_dtype, FLAGS.data_dir,
+                                                        train_split, copy.deepcopy(config.data))
+              return ds_lab_cls.input_fn(params)
+
+            fns[index] = _fn
+          return tf.switch_case(tf.cast(tf.train.get_global_step() // train_steps, tf.int32), fns)
```
fsx950223, line 483: Could you check whether the image size changes with the global step on TPU.

fsx950223:
Commit1 and commit2 are two solutions and you could try either one.

efficientnetv2/main.py
mingxingtan: Hmm, looks like it has some issues on TPU, maybe it is because TF1 doesn't do JIT compile and needs static shape at the first time of complication.

Could you help add it to main_tf2.py?  TF2 script can JIT compilation and can probably support this. I plan to make main_tf2.py as default.

fsx950223: It's a bug in switch_case(https://github.com/tensorflow/tensorflow/issues/49378), which happens on tf2 too.

fsx950223:
I updated the code and I'm sure it works.

fsx950223:
Does it work as expected?

fsx950223:
Strange, it's slower than original one.

fsx950223:
Test link https://colab.research.google.com/drive/1nxYvKnWPGMlmHTXSaU9wR8M7NI-7BoQJ?usp=sharing.

fsx950223:
Still don't support tpu

PR #1005, opened 2021-05-19
Add TF2 script and fix import issues.

mingxingtan:


mingxingtan:
@fsx950223 FYI, the tf2 script is partially based on the efficientdet-tf2 script.
It doesn't support ema yet, but it can still get comparable accuracy (top1=78.8%) with cosine decay for efficientnetv2-b0.

PR #1006, opened 2021-05-19
Add mode tf2mlir that exports MLIR

nihui:


fsx950223:
It's better to create another file called mlir.py and export mlir from saved model.

fsx950223 (approved):

PR #1011, opened 2021-05-22
fix test cases

fsx950223:


mingxingtan (approved):
Great, thanks for the fix!

PR #1025, opened 2021-06-11
fix parse_image_size

smedegaard:
`parse_image_size()` in `efficientdet/utils.py` swaps height and width of the image when you give a string as argument.

This PR fixes that and makes it possible to pass a `dict`. 

> Explicit is better than implicit.

### DIFF
--- a/efficientdet/.#utils.py
+++ b/efficientdet/.#utils.py
@@ -1,0 +1,1 @@
+anders@anders-pc.7481:1623328388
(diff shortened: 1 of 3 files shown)


smedegaard:
@googlebot I signed it!

smedegaard:
@googlebot I fixed it.

smedegaard:
@googlebot I fixed it.

PR #1193, opened 2023-05-13
p.add_(torch.sign(update), alpha=-group['lr']) has been made more eff…

sachinspanicker:
…icient by using the torch.sign function's inplace mode. This will prevent the need to create a new tensor for the updated parameter, which can save a significant amount of time for large models.

### DIFF
--- a/lion/lion_pytorch.py
+++ b/lion/lion_pytorch.py
@@ -77,7 +77,12 @@ class Lion(Optimizer):
 
         # Weight update
         update = exp_avg * beta1 + grad * (1 - beta1)
-        p.add_(torch.sign(update), alpha=-group['lr'])
+
+        p.add_(torch.sign(update), alpha=-group['lr'], inplace=True)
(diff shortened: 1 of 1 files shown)


crazydonkey200:
Sorry that I have missed this pull request earlier. Thanks for the contribution!

liuslnlp:
Error
```
add_() received an invalid combination of arguments - got unrecognized keyword arguments: inplace
```

My torch version
```
>>> import torch
>>> torch.__version__
'2.1.0.dev20230829+cu121'
```

fbotp:
@liuslnlp  `add_` is already an **inplace** method, so I remove this parameter in method, and it works, you can have a try

yongchanghao:
@crazydonkey200  As the comments above suggested, there isn't such an argument as `inplace=True` in `torch.Tensor.add_`. for both torch versions above or under 2.0. I believe this code hasn't been tested.

